PR #4, opened 2020-01-31
brought in changes from anomaly experiments

nishaq503:
fix: made things faster

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -1,13 +1,13 @@
-""" Clustered Hierarchical Entropy-scaling Manifold Mapping.
+""" Clustered Learning of Approximate Manifolds.
 
 # TODO: https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging
 """
+import concurrent.futures
 import logging
 import pickle
 import random
 from collections import deque
 from operator import itemgetter
-from queue import Queue
 from threading import Thread
 from typing import Set, Dict, Iterable, BinaryIO, List, Union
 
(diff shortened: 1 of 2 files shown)


thoward27:
@nishaq503 you broke a unit test. We cannot merge to master till that's addressed

pyclam/manifold.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, manifold: 'Manifold', argpoints: Vector, name: str, **kwargs)
         if not argpoints and self.children:
             self.argpoints = [p for child in self.children for p in child.argpoints]
         elif not argpoints:
-            raise ValueError(f"Cluster {name} needs argpoints.")
+            raise ValueError(f'Cluster {name} needs argpoints: {argpoints}')
```
thoward27: We already established there are no argpoints, logging them here seems weird

pyclam/manifold.py
```diff
@@ -308,18 +308,29 @@ def partition(self, *criterion) -> Iterable['Cluster']:
             self.manifold.data[farthest],
         ])
 
-        p1_idx, p2_idx = list(), list()
+        p1_idx, p2_idx = [self.argradius], [farthest]
         [(p1_idx if p1 < p2 else p2_idx).append(i)
          for batch in iter(self)
          for i, p1, p2 in zip(batch, *cdist(poles, self.manifold.data[batch], self.metric))]
 
+        # TODO: Neither of these should be possible
```
thoward27: Then why are we testing for them? bugs elsewhere?

pyclam/manifold.py
```diff
@@ -624,7 +649,7 @@ def build(self, *criterion) -> 'Manifold':
     def build_tree(self, *criterion) -> 'Manifold':
         """ Builds the Cluster-tree. """
         while True:
-            logging.info(f'current depth: {len(self.graphs) - 1}')
+            logging.info(f'current depth: {len(self.graphs) - 1}, {len(self.graphs[-1].clusters.keys())} clusters')
```
thoward27: Shouldn't need to access keys here

pyclam/manifold.py
nishaq503: I don't know why this invariant breaks. I'll look into it after unifying distance calls.

pyclam/manifold.py
nishaq503: addressed

pyclam/manifold.py
nishaq503: addressed

PR #5, opened 2020-01-31
unified distance calls

nishaq503:
feature: unified distance calls into single function

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -12,7 +12,7 @@ from threading import Thread
 from typing import Set, Dict, Iterable, BinaryIO, List, Union
 
 import numpy as np
-from scipy.spatial.distance import pdist, cdist
+from scipy.spatial.distance import cdist
 
 from pyclam.types import Data, Radius, Vector, Metric
 
@@ -49,6 +49,8 @@ class Cluster:
         self.manifold: 'Manifold' = manifold
         self.argpoints: Vector = argpoints
         self.name: str = name
+
+        self.distance = self.manifold.distance
 
         # TODO: Consider relying on Graph.edges instead of having neighbors be a member of Cluster.
         self.neighbors: Dict['Cluster', float] = dict()  # key is neighbor, value is distance to neighbor
@@ -132,10 +134,10 @@ class Cluster:
                 indices = self.argpoints
             else:
                 n = int(np.sqrt(len(self)))
-                indices = list(np.random.choice(self.argpoints, n, replace=False))
+                indices = [int(i) for i in np.random.choice(self.argpoints, n, replace=False)]
 
             # Handle Duplicates.
-            if pdist(self.manifold.data[indices], self.metric).max(initial=0.) == 0.:
+            if self.distance(indices, indices).max(initial=0.) == 0.:
                 indices = np.unique(self.manifold.data[self.argpoints], return_index=True, axis=0)[1]
                 indices = [self.argpoints[i] for i in indices][:n]
 
@@ -163,7 +165,7 @@ class Cluster:
         """ The index used to retrieve the medoid. """
         if '_argmedoid' not in self.__dict__:
             logging.debug(f"building cache for {self}")
-            _argmedoid = np.argmin(cdist(self.samples, self.samples, self.metric).sum(axis=1))
+            _argmedoid = np.argmin(self.distance(self.argsamples, self.argsamples).sum(axis=1))
             self.__dict__['_argmedoid'] = self.argsamples[int(_argmedoid)]
         return self.__dict__['_argmedoid']
 
@@ -188,7 +190,7 @@ class Cluster:
             logging.debug(f'building cache for {self}')
 
             def argmax_max(b):
-                distances = self.distance(self.manifold.data[b])
+                distances = self.distance(self.argmedoid, b)[0]
                 argmax = int(np.argmax(distances))
                 return b[argmax], distances[argmax]
 
@@ -207,7 +209,7 @@ class Cluster:
                 return 0.
             count = [d <= (self.radius / 2)
                      for batch in self
-                     for d in self.distance(self.manifold.data[batch])]
+                     for d in self.distance(self.argmedoid, batch)[0]]
             count = np.sum(count)
             self.__dict__['_local_fractal_dimension'] = count if count == 0. else np.log2(len(self.argpoints) / count)
         return self.__dict__['_local_fractal_dimension']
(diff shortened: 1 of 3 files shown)


pyclam/manifold.py
```diff
@@ -133,13 +135,15 @@ def argsamples(self) -> Vector:
             else:
                 n = int(np.sqrt(len(self)))
                 indices = list(np.random.choice(self.argpoints, n, replace=False))
+                indices = [int(i) for i in indices]
 
             # Handle Duplicates.
-            if pdist(self.manifold.data[indices], self.metric).max(initial=0.) == 0.:
+            if self.distance(x1=indices, x2=indices).max(initial=0.) == 0.:
```
thoward27: Recommendation to not use variable names here, this function *almost* justifies being position only. It doesn't matter which comes first.

pyclam/manifold.py
```diff
@@ -133,13 +135,15 @@ def argsamples(self) -> Vector:
...
                 indices = np.unique(self.manifold.data[self.argpoints], return_index=True, axis=0)[1]
                 indices = [self.argpoints[i] for i in indices][:n]
 
             # Cache it.
+            indices = [int(i) for i in indices]
```
thoward27: Why int here, they should be ints already, if they aren't we fucked up elsewhere

pyclam/manifold.py
```diff
@@ -133,13 +135,15 @@ def argsamples(self) -> Vector:
             else:
                 n = int(np.sqrt(len(self)))
                 indices = list(np.random.choice(self.argpoints, n, replace=False))
+                indices = [int(i) for i in indices]
```
thoward27: can't we do something like `random.choice(dtype=int)`

pyclam/manifold.py
```diff
@@ -163,7 +167,7 @@ def argmedoid(self) -> int:
         """ The index used to retrieve the medoid. """
         if '_argmedoid' not in self.__dict__:
             logging.debug(f"building cache for {self}")
-            _argmedoid = np.argmin(cdist(self.samples, self.samples, self.metric).sum(axis=1))
+            _argmedoid = np.argmin(self.distance(x1=self.argsamples, x2=self.argsamples).sum(axis=1))
```
thoward27: same here, would recommend against using keyword arguments for functions where the order doesn't matter

thoward27: https://www.python.org/dev/peps/pep-0570/

pyclam/manifold.py
```diff
@@ -207,7 +211,7 @@ def local_fractal_dimension(self) -> float:
                 return 0.
             count = [d <= (self.radius / 2)
                      for batch in self
-                     for d in self.distance(self.manifold.data[batch])]
+                     for d in self.distance(x1=self.argmedoid, x2=batch)[0]]
```
thoward27: why add the `[0]`

thoward27: this would indicate to me that the distance functions are materially different

pyclam/manifold.py
```diff
@@ -231,14 +235,14 @@ def tree_search(self, point: Data, radius: Radius, depth: int) -> Dict['Cluster'
 
         results: Dict['Cluster', Radius] = dict()
         if self.depth == depth:
-            results = {self: self.distance(np.asarray([point]))[0]}  # TODO: Cover
+            results = {self: self.distance(x1=self.argmedoid, x2=None, p2=point)[0]}  # TODO: Cover
```
thoward27: I definitely don't understand what p2 is intended to signify here

pyclam/manifold.py
```diff
@@ -258,8 +262,8 @@ def _tree_search(self, point: Data, radius: Radius, depth: int) -> Dict['Cluster
                 break
 
             # filter out clusters that are too far away to possibly contain any hits.
-            centers = np.asarray([c.medoid for c in children])
-            distances = cdist(np.expand_dims(point, 0), centers, self.metric)[0]
+            argcenters = [c.argmedoid for c in children]
+            distances = self.distance(x1=None, p1=point, x2=argcenters)[0]
```
thoward27: weird, I think this would read better if params were `x1, x2` and that's it.

pyclam/manifold.py
```diff
@@ -405,6 +398,9 @@ def __hash__(self):
     def __contains__(self, cluster: 'Cluster') -> bool:
         return cluster in self.clusters.keys()
 
+    def distance(self, *, x1: Union[int, List[int], None], x2: Union[int, List[int], None], p1: Data = None, p2: Data = None) -> np.ndarray:
+        return self.manifold.distance(x1=x1, x2=x2, p1=p1, p2=p2)
```
thoward27: Disagree with this implementation. I'd rather see just x1 x2. p1 p2 is super weird here for me

thoward27: https://www.python.org/dev/peps/pep-0570/

pyclam/manifold.py
```diff
@@ -608,6 +605,48 @@ def __repr__(self) -> str:
...
+
+        This DOES NOT do any batching.
+
+        The metric given to Manifold should have the following properties:
+            * dist(p1, p2) = 0 if and only if p1 = p2.
+            * dist(p1, p2) = dist(p2, p1)
+
+        :param x1: index of point or list of indexes of points in self.data.
+        :param x2: index of point or list of indexes of points in self.data.
+        :param p1: (optional) point(s) given directly to override x1. Useful when the point is not in self.data
```
thoward27: weird, I find this p1 p2 notion unintuitive, I'd say make it work with just x1 x2 and make them positional only

pyclam/manifold.py
```diff
@@ -608,6 +605,48 @@ def __repr__(self) -> str:
     def depth(self) -> int:
         return len(self.graphs) - 1
 
+    def distance(self, *, x1: Union[int, List[int], None], x2: Union[int, List[int], None], p1: Data = None, p2: Data = None) -> np.ndarray:
```
thoward27: I see now what you intended here, to override if point not in self.data, but this function should not do that. Call it `argdistance` if you want.

thoward27: perhaps implement checks on type, list of ints is indices, numpy array is a dedicated point. Something other than this param string

thoward27: Just had a thought.. check dimensions. 1D arrays are indices, 2D are points. State that in the docstring, and make this positional, that would be very satisfying

pyclam/manifold.py
nishaq503: I think numpy just maps that to np.int from native ints

pyclam/manifold.py
nishaq503: Line 143 makes it a list of numpy.int64

pyclam/manifold.py
```diff
@@ -608,6 +607,31 @@ def __repr__(self) -> str:
...
+        :param x1: a list of indices, or a 2D matrix of data points
+        :param x2: a list of indices, or a 2D matrix of data points
+        :return: matrix of pairwise distances.
+        """
+
+        x1, x2 = np.asarray(x1), np.asarray(x2)
+        
+        # Fetch data if given indices.
+        if len(x1.shape) < 2:
+            x1 = self.data[x1 if x1.ndim == 1 else np.expand_dims(x1, 0)]
```
nishaq503: This is wrong.
When x1.shape is less than 2, x1.ndim is always 1

thoward27: No, it can be zero. `len(np.array(1))`

pyclam/tests/test_cluster.py
```diff
@@ -182,7 +180,7 @@ def test_neighbors(self):
         return
 
     def test_distance(self):
-        self.assertGreater(self.children[0].distance(np.expand_dims(self.children[1].medoid, 0)), 0)
+        self.assertGreater(self.children[0].distance(x1=self.children[0].argmedoid, x2=self.children[1].argmedoid), 0)
```
thoward27: Drop arguments here

PR #9, opened 2020-02-04
added edge-pruniing

nishaq503:
added edge-pruning and ignoring extra-vacuous edges.

### DIFF
--- a/pyclam/datasets.py
+++ b/pyclam/datasets.py
@@ -30,7 +30,7 @@ def ring_data(n: int, radius: float, noise: float) -> np.ndarray:
     return np.asarray(ring, dtype=np.float64)
 
 
-def bullseye(n: int = 1_000, num_rings: int = 3, noise: float = 0.05) -> Tuple[np.ndarray, List[int]]:
+def bullseye(n: int = 2_000, num_rings: int = 3, noise: float = 0.05) -> Tuple[np.ndarray, List[int]]:
     data: np.ndarray = np.ndarray(shape=(0, 2))
     labels: List[int] = list()
     for r in range(1, 2 * num_rings, 2):
@@ -40,7 +40,7 @@ def bullseye(n: int = 1_000, num_rings: int = 3, noise: float = 0.05) -> Tuple[n
     return np.asarray(data, dtype=np.float64), labels
 
 
-def line(n: int = 1_000, m: float = 1, c: float = 0., noise: float = 0.05) -> Tuple[np.ndarray, List[int]]:
+def line(n: int = 5_000, m: float = 1, c: float = 0., noise: float = 0.05) -> Tuple[np.ndarray, List[int]]:
     x = np.random.rand(n)
     y = m * x + c
     data = np.asarray((x, y)).T
@@ -49,13 +49,13 @@ def line(n: int = 1_000, m: float = 1, c: float = 0., noise: float = 0.05) -> Tu
     return np.asarray(data, dtype=np.float64), list(labels)
 
 
-def xor(n: int = 1_000) -> Tuple[np.ndarray, List[int]]:
+def xor(n: int = 5_000) -> Tuple[np.ndarray, List[int]]:
     data = np.random.rand(n, 2)
     labels = [int((x > 0.5) != (y > 0.5)) for x, y, in data]
     return np.asarray(data, dtype=np.float64), labels
 
 
-def spiral_2d(n: int = 2_000, noise: float = 0.1) -> Tuple[np.ndarray, List[int]]:
+def spiral_2d(n: int = 5_000, noise: float = 0.1) -> Tuple[np.ndarray, List[int]]:
     theta = np.sqrt(np.random.rand(n)) * 2 * np.pi
 
     r_a = 2 * theta + np.pi
@@ -81,7 +81,7 @@ def generate_torus(n: int, r_torus: float, noise: float) -> Tuple[np.ndarray, np
     return x, y, z
 
 
-def tori(n: int = 7_500, noise: float = 0.015, r_torus: float = 1.) -> Tuple[np.ndarray, List[int]]:
+def tori(n: int = 10_000, noise: float = 0.015, r_torus: float = 1.) -> Tuple[np.ndarray, List[int]]:
     x, y, z = generate_torus(n=n // 2, r_torus=r_torus, noise=noise)
     torus_1 = np.stack([x - r_torus, y, z], axis=1)
     labels = [0 for _ in x]
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -1,6 +1,4 @@
 """ Clustered Learning of Approximate Manifolds.
-
-# TODO: https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging
 """
 import concurrent.futures
 import logging
@@ -9,7 +7,7 @@ import random
 from collections import deque
 from operator import itemgetter
 from threading import Thread
-from typing import Set, Dict, Iterable, BinaryIO, List, Union
+from typing import Set, Dict, Iterable, BinaryIO, List, Union, Tuple
 
 import numpy as np
 from scipy.spatial.distance import cdist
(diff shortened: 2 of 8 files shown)


pyclam/manifold.py
```diff
@@ -107,6 +107,11 @@ def depth(self) -> int:
         """ The depth in the tree at which the cluster exists. """
         return len(self.name)
 
+    @property
+    def parent(self) -> 'Cluster':
+        """ Returns the parent of self """
+        return self.manifold.select(self.name[:-1])
```
thoward27: Should we just store a parent reference?

pyclam/manifold.py
```diff
@@ -440,12 +444,40 @@ def _build_edges_matrix(self) -> None:
 
                 batch_left, batch_right = tuple(map(list, np.where(batch_differences <= 0.)))
                 [clusters[l_ + i].neighbors.update({clusters[r_]: batch_distances[l_, r_]}) for l_, r_ in zip(batch_left, batch_right) if (l_ + i) != r_]
+        return
+
+    def _build_edges_pruned(self) -> None:
```
thoward27: Does this pass unit tests that use it?

pyclam/manifold.py
```diff
@@ -440,12 +444,40 @@ def _build_edges_matrix(self) -> None:
...
+        root = self.manifold.select('')
+
+        for cluster in clusters:
+            ancestor, radius = cluster, cluster.radius
+            while radius <= 0.:
+                ancestor, radius = ancestor.parent, ancestor.parent.radius
+
+            potential_neighbors: Dict['Cluster', Radius] = root.tree_search(
+                point=cluster.medoid,
+                radius=radius * 2 * np.sqrt(2),
```
thoward27: Weird not seeing dimensionality used here

pyclam/manifold.py
nishaq503: We are not using this anymore. I took it out

pyclam/manifold.py
nishaq503: complexities cancel away

pyclam/manifold.py
nishaq503: It does now

pyclam/manifold.py
```diff
@@ -715,24 +716,26 @@ def _partition_threaded(self, criterion):
 
         return clusters
 
-    def select(self, name: str) -> Cluster:
```
thoward27: need a docstring here

pyclam/datasets.py
```diff
@@ -30,7 +30,7 @@ def ring_data(n: int, radius: float, noise: float) -> np.ndarray:
     return np.asarray(ring, dtype=np.float64)
 
 
-def bullseye(n: int = 1_000, num_rings: int = 3, noise: float = 0.05) -> Tuple[np.ndarray, List[int]]:
+def bullseye(n: int = 2_000, num_rings: int = 3, noise: float = 0.05) -> Tuple[np.ndarray, List[int]]:
```
thoward27: dataset changes shouldn't be part of this PR, not that it really matters, but try and keep them singular in purpose

pyclam/manifold.py
```diff
@@ -107,6 +107,9 @@ def depth(self) -> int:
         """ The depth in the tree at which the cluster exists. """
         return len(self.name)
 
+    def distance_from(self, x1: Union[List[int], Data]) -> np.ndarray:
```
thoward27: Clever, I like this one.

pyclam/manifold.py
```diff
@@ -280,6 +283,7 @@ def partition(self, *criterion) -> Iterable['Cluster']:
         2 children are produced if the cluster can be split, otherwise 1 child is produced.
         """
         if not all((
+                True if self.depth == 0 else self.name[-1] != '0',
```
thoward27: this reads weirdly, can it be done in two separate boolean clauses

pyclam/manifold.py
```diff
@@ -369,6 +375,7 @@ def __init__(self, *clusters):
 
         # self.clusters is a dictionary of the clusters in the graph and the connected component subgraph that the cluster belongs to.
         self.clusters: Dict[Cluster: 'Graph'] = {c: None for c in clusters}
+        self.count = 0
```
thoward27: What does this do

pyclam/manifold.py
```diff
@@ -410,42 +417,30 @@ def depth(self) -> int:
     def metric(self) -> Metric:
         return next(iter(self.clusters.keys())).metric
 
-    def _build_edges_matrix(self) -> None:
```
thoward27: love seeing this go

pyclam/manifold.py
```diff
@@ -410,42 +417,30 @@ def depth(self) -> int:
...
-
-                batch_distances = self.distance(batch_argcenters, argcenters)
-                batch_differences = ((batch_distances - radii).T - batch_radii).T
-
-                batch_left, batch_right = tuple(map(list, np.where(batch_differences <= 0.)))
-                [clusters[l_ + i].neighbors.update({clusters[r_]: batch_distances[l_, r_]}) for l_, r_ in zip(batch_left, batch_right) if (l_ + i) != r_]
-
-        return
-
     def build_edges(self) -> None:
```
thoward27: Is this the function that is only called at the bottom now? That should be specified if so

thoward27: This calls top-down, but I thought we discussed that the new function builds bottom-up, can you clarify?

pyclam/manifold.py
```diff
@@ -672,7 +673,7 @@ def build(self, *criterion) -> 'Manifold':
     def build_tree(self, *criterion) -> 'Manifold':
         """ Builds the Cluster-tree. """
         while True:
-            logging.info(f'current depth: {len(self.graphs) - 1}, {len(self.graphs[-1].clusters)} clusters')
+            logging.info(f'depth: {len(self.graphs) - 1}, {len(self.graphs[-1].clusters)} clusters')
```
thoward27: self.distance?

pyclam/manifold.py
```diff
@@ -715,24 +716,26 @@ def _partition_threaded(self, criterion):
...
-        cluster: Cluster = next(iter(self.graphs[0]))
+        root: Cluster = next(iter(self.graphs[0]))
+        lineage = [root]
         for depth in range(len(name) + 1):
-            partial_name = name[:depth]
-            for child in cluster.children:
-                if child.name == partial_name:
-                    cluster = child
+            for child in lineage[-1].children:
+                if child.name == name[:depth]:
```
thoward27: Should we make children of a cluster a set keyed on name?

pyclam/datasets.py
nishaq503: I kinda needed to to be able to measure any speedups

pyclam/manifold.py
nishaq503: Checking just self.name[-1] == '0' breaks at the root level.
I would rather just punt this until we find what is really wrong with partition that it produced the empty list that I check for in line 329

pyclam/manifold.py
nishaq503: ugh I've gotten rid of this line so many times

pyclam/manifold.py
nishaq503: It gets called once from the bottom layer. This should indeed be specified.
It iterates over the graphs from the graph at depth 1 to itself. This is handled in in lines 437-442.

Interestingly, the helper _find_neighbors is constructed in such a way so as not to repeat any distance computations in layers over which the graph was already build so it does not even matter if Graph.build_edges gets called multiple times per layer. It will only do the incremental work necessary to complete the building of any graphs.

pyclam/manifold.py
nishaq503: I don't get this

pyclam/manifold.py
nishaq503: On it.

pyclam/manifold.py
nishaq503: There are only one or two children so I do not see the change being worth it

pyclam/manifold.py
nishaq503: It was such common usage

pyclam/manifold.py
nishaq503: This is kinda equivalent to keeping a self.is_leaf flag

PR #10, opened 2020-02-08
Basic outrank algorithm

nishaq503:
Implemented the basic outrank algorithm.

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -3,10 +3,8 @@
 import concurrent.futures
 import logging
 import pickle
-import random
 from collections import deque
 from operator import itemgetter
-from threading import Thread
 from typing import Set, Dict, Iterable, BinaryIO, List, Union, Tuple
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


pyclam/manifold.py
```diff
@@ -486,33 +487,47 @@ def clear_cache(self) -> None:
...
-    def random_walk(self, steps: int = 5, walks: int = 1) -> Dict[Cluster, int]:
-        """ Performs a random walk, returning a modified graph instance.
-
-        :param int steps: number of steps per walk
-        :param int walks: number of walks to perform
-        :returns a Dict of cluster names to visit counts
-        """
-        clusters = list(self.clusters.keys())
-        results = {c: list() for c in clusters}
+    def _random_walks(self, clusters: List[Cluster], steps: int) -> Dict[Cluster, int]:
```
thoward27: Can this functionality be moved into `random_walks`? Let's keep namespace clean if we can since python doesn't provide true privates

PR #12, opened 2020-02-21
removed logging from manifold.py

nishaq503:
We need this so that:
* short term, I can make it super easy to reproduce our plots in the chaoda paper
* long term, we should not configure logging because overwriting the logging config is a python 3.8 only option and users will likely want it configure it themselves


thoward27:
Is there any safe way to configure logging that allows users to easily override? How do other libs handle this? It is hard to imagine no one out there is configuring loggers, or that everyone is preventing users from doing it as well.

PR #13, opened 2020-05-21
almost done

nishaq503:
Let's look this over and maybe write more tests


thoward27:
@nishaq503 I gave this a quick glance, nothing seems apparently long, but I would like you to talk me through the changes to expedite a deeper understanding

PR #19, opened 2020-06-14
enhancement: replace __dict with cache attribute

nishaq503:


pyclam/manifold.py
```diff
@@ -202,36 +202,37 @@ def argmax_max(b):
...
 
     def clear_cache(self) -> None:
         """ Clears the cache for the cluster. """
         logging.debug(f'clearing cache for {self}')
-        for prop in ['_argsamples', '_argmedoid', '_argradius', '_radius', '_local_fractal_dimension', '_optimal']:
-            try:
-                del self.__dict__[prop]
-            except KeyError:
-                pass
+        self.cache = {'optimal': False, 'absorbable': False}
```
thoward27, line 234: Seems weird to set values in clear

nishaq503: resets to defaults. See the init method.
In any case, I may have figured out how to not have those two flags in Cluster anyway

PR #20, opened 2020-06-14
refactor: Cluster naming scheme and Cluster.partition

nishaq503:
Cluster.partition can now truly deal with an arbitrary number of poles.
Cluster naming scheme has been updated to deal with arbitrary arity of the tree.

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -47,7 +47,7 @@ class Cluster:
         self.manifold: 'Manifold' = manifold
         self.argpoints: Vector = argpoints
         self.name: str = name
-        self.children: Set['Cluster'] = set()
+        self.children: Union[None, List['Cluster']] = None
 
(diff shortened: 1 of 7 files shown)


pyclam/manifold.py
```diff
@@ -87,7 +87,7 @@ def __str__(self) -> str:
 
     def __repr__(self) -> str:
         if 'repr' not in self.cache:
-            self.cache['repr'] = '-'.join([self.name, ', '.join(map(str, self.argpoints))])
+            self.cache['repr'] = '-'.join([self.name, ', '.join(map(str, sorted(self.argpoints)))])
```
thoward27, line 90: are we sure we want to sort the argpoints?

nishaq503: I like the visual, while debugging, of having indices increase as I read them from left to right

PR #26, opened 2020-08-26
SelectionClause for LR constants

nishaq503:
added ability to select clusters for optimal graph by using LR constants

### DIFF
--- a/pyclam/criterion.py
+++ b/pyclam/criterion.py
@@ -185,7 +185,7 @@ class SelectionClauses(SelectionCriterion):
         return selected
 
 
-class RegressionConstants(SelectionCriterion):
+class LinearRegressionConstants(SelectionCriterion):
     def __init__(self, constants: Union[np.array, List[float]]):
         constants = np.asarray(constants, dtype=float)
         if constants.shape != (6,):
--- a/pyclam/criterion.py
+++ b/pyclam/criterion.py
@@ -145,6 +145,7 @@ class SelectionClause(SelectionCriterion):
         self.min_radius, self.max_radius = float(radius_ratio[0]), float(radius_ratio[1])
 
     def __call__(self, root: Cluster) -> Set[Cluster]:
+        logging.info(f'selecting clusters for optimal graph...')
         manifold = root.manifold
         clusters: Set[Cluster] = {cluster for cluster in manifold.layers[1].clusters}
         selected: Set[Cluster] = set()
@@ -168,7 +169,8 @@ class SelectionClause(SelectionCriterion):
                     selected.add(cluster)
             clusters = new_clusters
 
-        return clusters
+        logging.info(f'selected {len(selected)} clusters for optimal graph')
+        return selected
 
 
 class LFDRange(SelectionCriterion):
--- a/pyclam/criterion.py
+++ b/pyclam/criterion.py
@@ -1,6 +1,6 @@
 import logging
 from abc import ABC, abstractmethod
-from typing import Set, Tuple, List, Iterable, Union, Dict
+from typing import Set, Tuple, List, Union, Dict
 
 import numpy as np
 from scipy.spatial.distance import cdist
@@ -111,29 +111,9 @@ class UniformDistribution(ClusterCriterion):
         return distance > 0.25
 
 
-class Clause(ClusterCriterion):
-    def __init__(
-            self,
-            lfd_ratio: Tuple[float, float] = (0, np.inf),
-            cardinality_ratio: Tuple[float, float] = (0, np.inf),
-            radius_ratio: Tuple[float, float] = (0, np.inf),
-    ):
-        self.min_lfd, self.max_lfd = float(lfd_ratio[0]), float(lfd_ratio[1])
-        self.min_cardinality, self.max_cardinality = float(cardinality_ratio[0]), float(cardinality_ratio[1])
-        self.min_radius, self.max_radius = float(radius_ratio[0]), float(radius_ratio[1])
-
-    def __call__(self, cluster: Cluster) -> bool:
-        lfd_ratio = cluster.local_fractal_dimension / cluster.parent.local_fractal_dimension
-        cardinality_ratio = cluster.cardinality / cluster.parent.cardinality
-        radius_ratio = max(cluster.radius, 1e-16) / max(cluster.parent.radius, 1e-16)
-        return all((
-                self.min_lfd <= lfd_ratio <= self.max_lfd,
-                self.min_cardinality <= cardinality_ratio <= self.max_cardinality,
-                self.min_radius <= radius_ratio <= self.max_radius,
-        ))
-
-
 class Leaves(SelectionCriterion):
+    """ Selects the leaves for building the graph.
+    """
     def __init__(self):
         return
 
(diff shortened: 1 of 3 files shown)


pyclam/criterion.py
```diff
@@ -111,6 +111,28 @@ def __call__(self, cluster: Cluster) -> bool:
         return distance > 0.25
 
 
+class Clause(ClusterCriterion):
```
thoward27, line 114: I would say `SelectionClause` may make this a bit clearer

thoward27, line 114: Also, docstrings

pyclam/criterion.py
```diff
@@ -119,6 +141,90 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
         return {cluster for cluster in root.manifold.layers[-1].clusters}
 
 
+class Layer(SelectionCriterion):
```
thoward27, line 144: Docstring

pyclam/criterion.py
```diff
@@ -119,6 +141,90 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+    def __init__(self, depth: int):
+        if depth <= 0:
+            raise ValueError(f'expected a positive depth. got: {depth}')
+        self.depth = depth
+
+    def __call__(self, root: Cluster):
```
thoward27, line 150: return type

pyclam/criterion.py
```diff
@@ -119,6 +141,90 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+        if manifold.depth <= self.depth:
+            return [cluster for cluster in manifold.layers[-1].clusters]
+        else:
+            return [cluster for cluster in manifold.layers[self.depth].clusters]
+
+
+class SelectionClauses(SelectionCriterion):
+    def __init__(self, clauses: Iterable[Clause]):
+        if any((not isinstance(c, Clause) for c in clauses)):
+            raise TypeError(f'selection clauses mut all be of type Clause.')
```
thoward27, line 161: typo

pyclam/criterion.py
```diff
@@ -119,6 +141,90 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+
+    def __call__(self, root: Cluster):
+        manifold: Manifold = root.manifold
+        if manifold.depth <= self.depth:
+            return [cluster for cluster in manifold.layers[-1].clusters]
+        else:
+            return [cluster for cluster in manifold.layers[self.depth].clusters]
+
+
+class SelectionClauses(SelectionCriterion):
```
thoward27, line 158: Docstring

pyclam/criterion.py
```diff
@@ -119,6 +141,90 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+
+        logging.info(f'selected {len(selected)} clusters for optimal graph')
+        return selected
+
+
+class LinearRegressionConstants(SelectionCriterion):
+    def __init__(self, constants: Union[np.array, List[float]]):
+        constants = np.asarray(constants, dtype=float)
+        if constants.shape != (6,):
+            raise ValueError(f'expected a vector of 6 elements. Got {constants.shape} instead.')
```
thoward27, line 192: I think 6 here should be a constant

pyclam/criterion.py
```diff
@@ -119,6 +141,90 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+                        new_clusters.update(cluster.children)
+                else:
+                    selected.add(cluster)
+            clusters = new_clusters
+
+        logging.info(f'selected {len(selected)} clusters for optimal graph')
+        return selected
+
+
+class LinearRegressionConstants(SelectionCriterion):
```
thoward27, line 188: Docstring

pyclam/manifold.py
```diff
@@ -81,10 +81,7 @@ def __eq__(self, other: 'Cluster') -> bool:
 
     def __lt__(self, other: 'Cluster') -> bool:
         """ For sorting clusters in Graphs. Sorts by depth, breaking ties by name. """
-        if self.depth == other.depth:
-            return self.name < other.name
-        else:
-            return self.depth < other.depth
+        return (self.name < other.name) if self.depth == other.depth else (self.depth < other.depth)
```
thoward27, line 84: image

pyclam/manifold.py
```diff
@@ -112,10 +109,12 @@ def __contains__(self, point: Data) -> bool:
 
     @property
     def parent(self) -> 'Cluster':
-        if self.depth > 0:
-            return self.manifold.ancestry(self)[-2]
-        else:
-            raise ValueError(f"root cluster has no parent")
+        if 'parent' not in self.cache:
```
thoward27, line 112: Should we just start spawning children with a `parent` attribute?

pyclam/manifold.py
```diff
@@ -224,6 +223,9 @@ def radius(self) -> Radius:
         if 'radius' not in self.cache:
             logging.debug(f'building cache for {self}')
             _ = self.argradius
+            if self.cache['radius'] < 0:
+                raise ValueError(f'got cluster {self.name} with negative radius. '
+                                 f'Make sure that the distance function used always returns non-negative values.')
```
thoward27, line 228: I think this check belongs wherever we set `self.cache['radius']`

pyclam/manifold.py
```diff
@@ -232,14 +234,52 @@ def local_fractal_dimension(self) -> float:
...
-            self.cache['local_fractal_dimension'] = count if count == 0. else np.log2(len(self.argpoints) / count)
+                self.cache['local_fractal_dimension'] = 1.
+            else:
+                count = sum([
+                    1 if distance <= (self.radius / 2) else 0
+                    for batch in iter(self)
+                    for distance in self.distance_from(batch)
+                ])
+                self.cache['local_fractal_dimension'] = 1. if count == 0 else np.log2(self.cardinality / count)
         return self.cache['local_fractal_dimension']
```
thoward27, line 245: When can we drop support for 3.6 and just use `functools.cached_property`

pyclam/manifold.py
```diff
@@ -232,14 +234,52 @@ def local_fractal_dimension(self) -> float:
...
+        """
+        return self.manifold.ancestry(self)[:-1]
+
+    @property
+    def descendents(self) -> List['Cluster']:
+        """ All clusters in the subtree starting at self, excluding self.
+        """
+        return [cluster for layer in self.manifold.layers[self.depth + 1:]
+                for cluster in layer.clusters
+                if self.name == cluster.name[:len(self.name)]]
```
thoward27, line 259: Doesn't this only get the next layer? And why not just access children? In any event, it looks like docstring != implementation

pyclam/manifold.py
```diff
@@ -412,7 +452,7 @@ def __eq__(self, other: 'Graph') -> bool:
                 set(self.edges[cluster]) == set(other.edges[cluster])
                 for cluster in self.clusters
             ))
-        return cluster_equality and edges_equality
+            return cluster_equality and edges_equality
```
thoward27, line 455: looks like weird spacing in the diff

pyclam/manifold.py
```diff
@@ -424,7 +464,7 @@ def __iter__(self) -> Iterable[Cluster]:
     def __str__(self) -> str:
         # Cashing value because sort can be expensive on many clusters.
         if 'str' not in self.cache:
-            self.cache['str'] = ','.join(sorted(list(map(str, self.clusters))))
+            self.cache['str'] = ', '.join(sorted(list(map(str, self.clusters))))
```
thoward27, line 467: I'm becoming a big fan of the `::` separator, I submit that for your consideration here

pyclam/criterion.py
nishaq503: removed this decision-tree based criterion

pyclam/criterion.py
nishaq503: done

pyclam/criterion.py
nishaq503: done

pyclam/criterion.py
nishaq503: removed this decision-tree based criterion

pyclam/criterion.py
nishaq503: we have been unable to come up with more features to use for the meta-ml models. If we ever do, we will change 6 as needed

pyclam/manifold.py
nishaq503: done. see Cluster.partition

pyclam/manifold.py
nishaq503: When we switch to a Rust backend, we will use proper caching. For now, I don't really want to drop python 3.6

pyclam/manifold.py
nishaq503: I did that because of some Rust instincts translating over into my python. The tab makes that a proper, exhaustive if-else statement.

pyclam/manifold.py
nishaq503: meh. I was looking at clusters in the debugger and the space helped visually.

pyclam/manifold.py
nishaq503: line 258. that's a list slice at the end. next layer to the final layer

Doing this using children would require a recursive approach and takes too long in python.

pyclam/manifold.py
nishaq503: moved

PR #28, opened 2020-10-25
Feature/manifold dot file

ali-amani01:
The function get_dot_file_string() has been added to the manifold class in Manifold.py, which returns a string that can be written on a dot file to produce a graph of the manifold. There are edges between parent and child clusters, as well as edges between overlapping clusters that are on the same level of the manifold.

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -5,7 +5,7 @@ import logging
 import pickle
 from collections import deque
 from operator import itemgetter
-from typing import Set, Dict, Iterable, BinaryIO, List, Union, Tuple, IO, Any
+from typing import Set, Dict, Iterable, BinaryIO, List, Union, Tuple, IO, Any, Optional, Callable
 
 import numpy as np
 from scipy.spatial.distance import cdist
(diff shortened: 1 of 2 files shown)


pyclam/manifold.py
```diff
@@ -24,12 +24,10 @@
 
 class Cluster:
     """ A cluster of points.
-
```
thoward27, line 27: These formatting changes seem unwanted

pyclam/manifold.py
```diff
@@ -115,6 +112,13 @@ def parent(self) -> 'Cluster':
                 raise ValueError(f"root cluster has no parent")
         return self.cache['parent']
 
+    @property
+    def color(self) -> str:
+        """ The color of the cluster node, derived from its LFD """
+        x = 29 + ((19 * 2) * int(round(self.local_fractal_dimension)))
+        y = 206 - ((22 * 2) * int(round(self.local_fractal_dimension)))
+        return '#' + '%02x%02x%02x' % (x, 0, y)
```
thoward27, line 120: I would prefer to see f-strings.

pyclam/manifold.py
```diff
@@ -115,6 +112,13 @@ def parent(self) -> 'Cluster':
                 raise ValueError(f"root cluster has no parent")
         return self.cache['parent']
 
+    @property
+    def color(self) -> str:
+        """ The color of the cluster node, derived from its LFD """
+        x = 29 + ((19 * 2) * int(round(self.local_fractal_dimension)))
+        y = 206 - ((22 * 2) * int(round(self.local_fractal_dimension)))
```
thoward27, line 119: These look like magic lines, should have some comments

pyclam/manifold.py
```diff
@@ -1121,7 +1115,6 @@ def __init__(
             **kwargs
     ):
         """ A Manifold needs the data from which to learn the manifold, and a distance function to use while doing so.
-
```
thoward27, line 1124: Very weird seeing so many formatting changes in docstrings. I would like to see these changes reverted

pyclam/manifold.py
```diff
@@ -1181,15 +1174,56 @@ def clear_cache(self):
     def depth(self) -> int:
         return len(self.layers) - 1
 
+    def get_dot_file_string(self, graph_name):
+        """
+        checks over every cluster in the manifold, finding their parent edges and all edges between two clusters on the
+        same level, and then writes the information for each cluster and each edge onto a string that can be written
+        on a dot file
+        """
```
thoward27, line 1182: bad docstring

thoward27 (changes requested):
@ali-amani01 I saw a notification for this PR and thought I'd lend a hand with the review, I hope that's okay with you and @nishaq503. The first pass looked OK, but I do have some concerns about merging this in as-is.

pyclam/manifold.py
nishaq503: These numbers need explanation. I would actually prefer if ```color``` were a passable function to the dot-file method. That way this code would live on the side of the user and the user would have complete control over colors for many different applications..

pyclam/manifold.py
nishaq503: f-strings are our preferred default for string formatting. They are much easier to read, and are faster in execution.

pyclam/manifold.py
```diff
@@ -1181,15 +1174,56 @@ def clear_cache(self):
     def depth(self) -> int:
         return len(self.layers) - 1
 
+    def get_dot_file_string(self, graph_name):
```
nishaq503, line 1177: ```graph_name``` does not need to be an argument. ```str(self)``` will get you the same thing

nishaq503, line 1177: This method should live in the Graph class

pyclam/manifold.py
```diff
@@ -1181,15 +1174,56 @@ def clear_cache(self):
...
+        checks over every cluster in the manifold, finding their parent edges and all edges between two clusters on the
+        same level, and then writes the information for each cluster and each edge onto a string that can be written
+        on a dot file
+        """
+        # starting the graph string with the graph name (an inputted value), and the root cluster
+        dot_file_string = 'graph ' + str(graph_name) + '{ \n' +\
+                          f'	edge[style=solid, penwidth="5",labeldistance="10"]\n' \
+                          f'	{self.root}[label="NODE: {self.root.name}\\nLFD: ' \
+                          f'{round(self.root.local_fractal_dimension, 3)}", ' \
+                          f'color="{self.root.color}", style=filled]\n'
```
nishaq503, line 1188: When using f-strings, you don't need to use + for concatenation  nor do you need \ for line breaks

pyclam/manifold.py
```diff
@@ -1181,15 +1174,56 @@ def clear_cache(self):
...
+
+        # an array of clusters that were already written to the dot file
+        visited_clusters = []
```
nishaq503, line 1191: This should be a Set rather than a list. Sets have faster lookups and appends. These are the only things you are using this object for. ```Set.add(item)``` is the analog of ```List.append(item)```

pyclam/manifold.py
```diff
@@ -1181,15 +1174,56 @@ def clear_cache(self):
...
+        # reading each layer of the manifold after the root
+        for i in range(1, self.depth + 1):
+            layer = self.layers[i]
+            edges = layer.build_edges().edges
+            for cluster in layer:
+                if cluster not in visited_clusters:
+
+                    # writing the current cluster, if it hasn't been visited already, to the string, and then writing
+                    # the edge between the cluster and its parent
+                    dot_file_string += f'	{cluster}[label="NODE: {cluster.name}\\nLFD: ' \
```
nishaq503, line 1202: I believe this loop could be a lot faster. Let's do a zoom call about the thought process here

pyclam/manifold.py
```diff
@@ -1267,7 +1301,6 @@ def _partition_threaded(self, criterion) -> List[Cluster]:
 
     def ancestry(self, cluster: Union[str, Cluster]) -> List[Cluster]:
         """ Returns the sequence of clusters that needs to be traversed to reach the requested cluster.
-
```
nishaq503, line 1270: Please undo the changes to the Doc-String formatting. We used the format recommended in PEP so that our doc-strings can be used to automatically generate documentation

nishaq503 (changes requested):
I have suggested some changes. Let's do a zoom call for that loop. I'm concerned that it will take too long in execution on large graphs.

PR #29, opened 2020-11-02
integrating the chaoda methods as a module

nishaq503:


pyclam/criterion.py
```diff
@@ -130,38 +130,89 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+
+
 class LinearRegressionConstants(SelectionCriterion):
     """ Uses constants from a meta-ml model using linear regression.
     There are 6 constants, one for each of the ratios and ema_ratios in Cluster.
     """
 
-    def __init__(self, constants: Union[np.array, List[float]]):
-        num_constants = 6
+    def __init__(self, constants: Union[np.array, List[float]], *, mode: str = 'ranked'):
```
thoward27, line 202: docstring should specify what modes are allowed

pyclam/criterion.py
```diff
@@ -130,38 +130,89 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+        if mode not in SELECTION_MODES.keys():
+            raise ValueError(f'mode must be one of {list(SELECTION_MODES.keys())}. Gor {mode} instead.')
 
+        num_constants = 6
         constants = np.asarray(constants, dtype=float)
         if constants.shape != (num_constants,):
             raise ValueError(f'expected a vector of {num_constants} elements. Got {constants.shape} instead.')
         self.constants: np.array = constants
+        self.mode = SELECTION_MODES[mode]
```
thoward27, line 211: I would set this right after checking the validity of mode

pyclam/criterion.py
```diff
@@ -176,6 +227,29 @@ def _predict_auc(self, manifold: Manifold) -> Dict[Cluster, float]:
         return predicted_auc
 
 
+class RegressionTreePaths(SelectionCriterion):
+    """ Uses the decision paths of the Regression Tree meta-ml models to select best clusters for graph.
+    """
+    def __init__(self, predict_auc: Callable[[Cluster], float], *, mode: str = 'ranked'):
+        if mode not in ['ranked', 'percentile']:
+            raise ValueError(f'mode must be either \'ranked\' or \'percentile\'. Gor {mode} instead.')
```
thoward27, line 235: typo

pyclam/criterion.py
```diff
@@ -176,6 +227,29 @@ def _predict_auc(self, manifold: Manifold) -> Dict[Cluster, float]:
         return predicted_auc
 
 
+class RegressionTreePaths(SelectionCriterion):
+    """ Uses the decision paths of the Regression Tree meta-ml models to select best clusters for graph.
+    """
+    def __init__(self, predict_auc: Callable[[Cluster], float], *, mode: str = 'ranked'):
```
thoward27, line 233: should include modes in docstring somewhere

pyclam/chaoda.py
```diff
@@ -0,0 +1,148 @@
+from typing import List, Dict, Set
```
thoward27, line 1: Module level docstring describing what this file provides

pyclam/chaoda.py
```diff
@@ -0,0 +1,148 @@
...
+            scores = erf((scores - mu) / (sigma * np.sqrt(2)))
+        elif mode == 'sigmoid':
+            scores = 1 / (1 + np.exp(-(scores - mu) / sigma))
+        else:
+            raise ValueError(f'normalization mode {mode} is undefined. Use one of {_NORMALIZATION_MODES}.')
+
+    return scores.ravel().clip(0, 1)
+
+
+def cluster_cardinality(graph: Graph, normalize: str = 'gaussian') -> np.array:
```
thoward27, line 33: docstring should provide available normalization modes

thoward27 (changes requested):
Caught a typo and some docstring format issues. 

Other than that, it appears that this PR should be fine to merge in, I had assumed this one would contain the removal of the graph random walk method, which it doesn't. I do wonder if this is the best way to package the chaoda work into clam. Is the idea here that a user would create a manifold using clam, then induce a graph using a selection criteria, then pass that graph to methods in chaoda? I wonder if we can simplify this process for our users, perhaps we could inherent from Manifold in chaoda to create an anomaly detection class that can be instantiated with data and used immediately?

I also wonder if we should change the name of the file all together, from chaoda to something like anomalies, something more explicit in its purpose.

pyclam/chaoda.py
```diff
@@ -0,0 +1,148 @@
...
+    scores: Dict[int, int] = {p: -cluster.cardinality for cluster in graph.clusters for p in cluster.argpoints}
+    scores: List[int] = [scores[i] for i in range(len(scores))]
+    return _normalize(np.asarray(scores, dtype=float), normalize)
+
+
+def parent_child(graph: Graph, normalize: str = 'gaussian') -> np.array:
+    """
+    Determines outlier scores for points by considering ratios of cardinalities of parent-child clusters.
+    The ratios are weighted by the child's depth in the tree, and are then accumulated for each point in each cluster in the graph.
+    Points with relatively low accumulated ration are the outliers.
```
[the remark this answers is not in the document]
nishaq503: fixed

pyclam/criterion.py
nishaq503: fixed

pyclam/criterion.py
nishaq503: fixed

pyclam/criterion.py
nishaq503: fixed

pyclam/criterion.py
nishaq503: fixed

pyclam/chaoda.py
nishaq503: fixed

pyclam/chaoda.py
nishaq503: fixed

pyclam/chaoda.py
```diff
@@ -0,0 +1,155 @@
...
+        for i in range(1, len(ancestry)):
+            score = float(ancestry[i-1].cardinality) / (ancestry[i].cardinality * np.sqrt(i))
+            for p in cluster.argpoints:
+                results[p] += score
+    return _normalize(results, normalize)
+
+
+def graph_neighborhood(graph: Graph, normalize: str = 'gaussian') -> np.array:
+    """
+    Determines outlier scores by the considering the relative graph-neighborhood of clusters.
```
thoward27, line 92: Docstrings should start on the line with the `"""`

pyclam/chaoda.py
```diff
@@ -0,0 +1,155 @@
...
+    Parameters
+    ----------
+    graph: Graph on which to calculate outlier scores.
+    normalize: which normalization mode to use to get scores in a [0, 1] range.
+               Must be one of 'linear', 'gaussian', or 'sigmoid'.
+
+    Returns
+    -------
+    A numpy array of outlier scores for each point in the manifold that the graph belongs to.
+
```
thoward27, line 50: I think we should be consistent with our docstrings, these should match the other docstrings in clam

thoward27 (commented):
Not sure how strongly you feel about the new format for docstrings, I'd prefer if we could be consistent with the code already in clam

pyclam/chaoda.py
nishaq503: reverted to old format

pyclam/chaoda.py
```diff
@@ -0,0 +1,129 @@
...
+    else:
+        mu: float = float(np.mean(scores))
+        sigma: float = max(float(np.std(scores)), 1e-3)
+
+        if mode == 'gaussian':
+            scores = erf((scores - mu) / (sigma * np.sqrt(2)))
+        elif mode == 'sigmoid':
+            scores = 1 / (1 + np.exp(-(scores - mu) / sigma))
+        else:
+            raise ValueError(f'normalization mode {mode} is undefined. Use one of {_NORMALIZATION_MODES}.')
```
thoward27, line 29: I am surprised you had to implement this, if there's a pre-built scikit implementation I think we should use that over re-implementing it here

pyclam/chaoda.py
```diff
@@ -0,0 +1,129 @@
...
+
+    :param graph: Graph on which to calculate outlier scores.
+    :param normalize: which normalization mode to use to get scores in a [0, 1] range.
+                      Must be one of 'linear', 'gaussian', or 'sigmoid'.
+    :return: A numpy array of outlier scores for each point in the manifold that the graph belongs to.
+    """
+    if normalize not in _NORMALIZATION_MODES:
+        raise ValueError(f'Normalization method {normalize} is undefined. Must by one of {_NORMALIZATION_MODES}.')
+
+    scores: Dict[int, int] = {p: -cluster.cardinality for cluster in graph.clusters for p in cluster.argpoints}
```
thoward27, line 47: These two lines are a bit odd to parse, is the first line intended to filter out duplicate points in the graph?

pyclam/chaoda.py
```diff
@@ -0,0 +1,129 @@
...
+    The ratios are weighted by the child's depth in the tree, and are then accumulated for each point in each cluster in the graph.
+    Points with relatively low accumulated ration are the outliers.
+
+    :param graph: Graph on which to calculate outlier scores.
+    :param normalize: which normalization mode to use to get scores in a [0, 1] range.
+                      Must be one of 'linear', 'gaussian', or 'sigmoid'.
+    :return: A numpy array of outlier scores for each point in the manifold that the graph belongs to.
+    """
+    if normalize not in _NORMALIZATION_MODES:
+        raise ValueError(f'Normalization method {normalize} is undefined. Must by one of {_NORMALIZATION_MODES}.')
```
thoward27, line 64: This could be good to extract as something like `trap_normalization` or `assert_normalization`

pyclam/chaoda.py
```diff
@@ -0,0 +1,129 @@
...
+        frontier: Set[Cluster] = {start}
+        while frontier:
+            visited += frontier
+            path_length.append(len(visited))
+            frontier = {neighbor for cluster in frontier for neighbor in graph.neighbors(cluster)}
+            frontier -= visited
+        return path_length
+
+    path_lengths: Dict[Cluster, List[int]] = {cluster: _bft(cluster) for cluster in graph.clusters}
+    scores: Dict[Cluster, int] = {cluster: path_length[len(path_lengths) // 4] if len(path_lengths) > 0 else 0 for cluster, path_length in path_lengths.items()}
```
thoward27, line 102: Why divided by 4?

pyclam/criterion.py
```diff
@@ -130,38 +130,103 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+    rankings: Dict[Cluster, float] = {cluster: 0 for cluster in predicted_auc.keys()}
+    for leaf in manifold.layers[-1].clusters:
+        # sort ancestors by predicted auc in non-increasing order, break ties by smaller depth
+        ancestors = list(sorted(manifold.ancestry(leaf), key=lambda c: (predicted_auc[c], c), reverse=True))
+        # add index of cluster in sorted list of ancestors to rankings
+        for i, cluster in enumerate(ancestors):
+            rankings[cluster] += (i * cluster.cardinality)
+    rankings.pop(manifold.select(''))  # remove root from rankings because ratios are undefined
+
+    # TODO: misses some points/clusters. breaks graph invariant
```
thoward27, line 165: is this TODO still true?

pyclam/criterion.py
```diff
@@ -176,6 +241,30 @@ def _predict_auc(self, manifold: Manifold) -> Dict[Cluster, float]:
...
+            raise ValueError(f'mode must be either \'ranked\' or \'percentile\'. Got {mode} instead.')
+        self.mode = SELECTION_MODES[mode]
+        self.predict_auc: Callable[[Cluster], float] = predict_auc
+
+    def __call__(self, root: Cluster) -> Set[Cluster]:
+        predicted_auc: Dict[Cluster, float] = {
+            cluster: self.predict_auc(cluster)
+            for layer in root.manifold.layers
+            for cluster in layer.clusters
+            if cluster.depth == layer.depth
```
thoward27, line 263: Could this ever be false?

thoward27 (approved):
I had a few questions, I think there are a few places where comments could help, but overall, I think I'm fine merging this in as-is now.

pyclam/chaoda.py
nishaq503: I know that 'linear' is in sklearn as 'min-max normalization' and I know that 'gaussian' is not in there. I decided to just implement the ones we need. We default to 'gaussian' because it produced the best results.

pyclam/chaoda.py
nishaq503: There are no duplicates in the graph. That's part of the Graph Invariant. Scores are supposed to be for points rather than for clusters. there are repeated scores, i.e. values, but not repeated keys, i.e. point indices.

pyclam/chaoda.py
nishaq503: You're right. I'll add that when I make the random-walks PR

pyclam/chaoda.py
nishaq503: a quarter of the max-path-length. It worked fine. We could learn the number based on the trends of how path-lengths increase but I did not want to tackle that complexity right now. I haven't completely thought through it

pyclam/criterion.py
nishaq503: I haven't checked since the Graph refactor. There was an insidious, hidden bug in the old version of Graph that showed up in all sorts of places. I hope this is fixed now. I'll check in the other PR

pyclam/criterion.py
nishaq503: Yes. Due to leaves from previous layers

PR #31, opened 2020-11-06
feat: integrated CHESS module

nishaq503:
Created a CHESS module and moves all search functionality out of CLAM.

I think we should stick with this class, rather than using mixins or inheriting from Manifold, because of the upcoming Rust implementation. We will first implement CLAM in Rust, and start to use rust-clam in the chaoda and chess modules as they are in python. Later, when we implement the chess and chaoda modules in Rust, the current design will be much easier to port over in manageable chunks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "clam"
-version = "0.2.9"
+version = "0.2.10"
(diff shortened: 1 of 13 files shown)


thoward27:
@nishaq503 why did we close this?

PR #34, opened 2020-11-09
feat: eccentricity, diameter, and as-matrix methods in Graph

nishaq503:
feat: usage of new methods in CHAODA

We now sit on 6 CHAODA methods, potentially 7 after today's meeting.
Later, I might add a faster estimate for Graph.diameter

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "clam"
-version = "0.2.10"
+version = "0.1.0"
 authors = ["Tom Howard <info@tomhoward.codes>", "Najib Ishaq <nishaq@my.uri.edu>", "Noah Daniels <noah_daniels@uri.edu>"]
 edition = "2018"
 
--- a/pyclam/chaoda.py
+++ b/pyclam/chaoda.py
@@ -88,19 +88,19 @@ def graph_neighborhood(graph: Graph, normalize: str = 'gaussian') -> np.array:
     """
     _catch_normalization_mode(normalize)
 
-    def _bft(start: Cluster) -> List[int]:
-        path_length: List[int] = list()
+    def _neighborhood_size(start: Cluster, steps: int) -> int:
+        """ Returns the number of clusters within 'steps' of 'start'. """
         visited: Set[Cluster] = set()
         frontier: Set[Cluster] = {start}
-        while frontier:
-            visited.update(frontier)
-            path_length.append(len(visited))
-            frontier = {neighbor for cluster in frontier for neighbor in graph.neighbors(cluster)}
-            frontier -= visited
-        return path_length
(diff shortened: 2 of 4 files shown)


pyclam/chaoda.py
```diff
@@ -217,4 +216,23 @@ def random_walk(graph: Graph, normalize: str = 'gaussian') -> np.array:
     return _normalize(np.array([scores[i] for i in range(graph.population)], dtype=int), mode=normalize)
 
 
+def stationary_probabilities(graph: Graph, normalize: str = 'gaussian') -> np.array:
+    scores: Dict[Cluster, float] = {cluster: -1 for cluster in graph.clusters}
```
thoward27, line 220: docstring

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "clam"
-version = "0.2.10"
```
thoward27, line 3: I just realized, we don't need to update the rust toml with pyclam, as the python code has no way of being included in the eventual rust crate. We could reset this to 0.1.0 and tie bumps to this to correspond with actual rust releases. The version numbers would be misaligned, but I think that's more informative then this linking we've done so far.

thoward27: Basically, I'd argue the python and rust implementations *should* be on different version numbers to indicate the relative progress both packages have had

pyclam/chaoda.py
```diff
@@ -88,19 +88,18 @@ def graph_neighborhood(graph: Graph, normalize: str = 'gaussian') -> np.array:
     """
     _catch_normalization_mode(normalize)
 
-    def _bft(start: Cluster) -> List[int]:
-        path_length: List[int] = list()
+    def _bft(start: Cluster, steps: int) -> int:
```
thoward27, line 91: seems odd to require a set amount of steps for bft

thoward27 (changes requested):
requesting changes because of a missing docstring on a public function. Other than that I had a thought about our versioning scheme and a nit

Cargo.toml
nishaq503: I agree with this. Let's start this with the CHESS branch.

pyclam/chaoda.py
nishaq503: it returns the number of clusters within a graph distance of 'steps' from 'start'. I changed the name and added a docstring

pyclam/chaoda.py
nishaq503: done

thoward27 (approved):
LGTM

thoward27:
@nishaq503 I still think this PR should *reset* the cargo.toml version to 0.1.0 to indicate the progress we've made there

PR #35, opened 2020-11-10
Methods to add and/or replace optimal graphs in Manifold

nishaq503:
feat: manifold.add_graphs and replace_graphs methods
fix: clearing the cache of Manifold after build methods

I'm experimenting with repeated builds of optimal graphs in chaoda to try and iteratively improve the learned selection criteria

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -948,12 +948,10 @@ class Manifold:
         self.root.candidates = {self.root: 0.}
 
         if len(selection_criteria) > 0:
-            logging.info(f'building graphs with {len(selection_criteria)} Selection Criteria.')
-            self.graphs = [Graph(*select(self.root)).build_edges() for select in selection_criteria]
+            self.add_graphs(*selection_criteria)
         else:
             logging.warning('No Selection Criterion was provided. Using leaves for building graph.')
             self.graphs = [Graph(*[cluster for cluster in self.layers[-1].clusters]).build_edges()]
-
         return self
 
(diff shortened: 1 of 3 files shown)


pyclam/manifold.py
```diff
@@ -965,8 +965,31 @@ def build_tree(self, *criterion) -> 'Manifold':
...
+    def add_graphs(self, *criterion) -> 'Manifold':
+        """ Uses the given selection criteria to add more graphs to the Manifold. """
+        from pyclam.criterion import SelectionCriterion
+
+        if not all((isinstance(c, SelectionCriterion) for c in criterion)):
+            raise ValueError(f'Only Selection Criteria are allowed for building graphs.')
+
+        logging.info(f'building graphs with {len(criterion)} Selection Criteria.')
+        self.graphs.extend([Graph(*select(self.root)).build_edges() for select in criterion])
+        self.cache.clear()
```
thoward27, line 980: I'm currently against clearing caches by default this way. What's the argument for forcibly clearing them at the end? I'd almost rather see it at the front, or just selectively clearing what it needs to work.

pyclam/manifold.py
```diff
@@ -965,8 +965,31 @@ def build_tree(self, *criterion) -> 'Manifold':
...
         return self
 
+    def replace_graphs(self, *criterion) -> 'Manifold':
```
thoward27, line 983: this seems like a good function to clear cache

thoward27 (commented):
Had some concerns about cache management

PR #36, opened 2020-11-11
utils file for common functions and constants

nishaq503:
feat: broke out utils file for common usage

### DIFF
--- a/pyclam/chaoda.py
+++ b/pyclam/chaoda.py
@@ -4,66 +4,40 @@
 from typing import List, Dict, Set
 
 import numpy as np
-from scipy.special import erf
 
 from pyclam import Graph, Cluster, Edge
+from pyclam.utils import catch_normalization_mode, normalize
 
 
-def _catch_normalization_mode(mode: str) -> None:
-    modes: List[str] = ['linear', 'gaussian', 'sigmoid']
-    if mode not in modes:
-        raise ValueError(f'Normalization method {mode} is undefined. Must by one of {modes}.')
-    else:
-        return
-
-
-def _normalize(scores: np.array, mode: str) -> np.array:
-    if mode == 'linear':
-        min_v, max_v, = float(np.min(scores)), float(np.max(scores))
-        if min_v == max_v:
-            max_v += 1.
-        scores = (scores - min_v) / (max_v - min_v)
-    else:
-        mu: float = float(np.mean(scores))
-        sigma: float = max(float(np.std(scores)), 1e-3)
-
-        if mode == 'gaussian':
-            scores = erf((scores - mu) / (sigma * np.sqrt(2)))
-        else:
-            scores = 1 / (1 + np.exp(-(scores - mu) / sigma))
-
-    return scores.ravel().clip(0, 1)
-
-
-def cluster_cardinality(graph: Graph, normalize: str = 'gaussian') -> np.array:
+def cluster_cardinality(graph: Graph, normalization_mode: str = 'gaussian') -> np.array:
     """ Determines outlier scores for points by considering the relative cardinalities of the clusters in the graph.
 
     Points in clusters with relatively low cardinalities are the outliers.
 
     :param graph: Graph on which to calculate outlier scores.
-    :param normalize: which normalization mode to use to get scores in a [0, 1] range.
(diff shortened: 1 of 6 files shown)


pyclam/chaoda.py
```diff
@@ -214,21 +188,21 @@ def random_walk(graph: Graph, normalize: str = 'gaussian') -> np.array:
 
     # normalize counts
     scores: Dict[int, int] = {point: -count for cluster, count in visit_counts.items() for point in cluster.argpoints}
-    return _normalize(np.array([scores[i] for i in range(graph.population)], dtype=int), mode=normalize)
+    return normalize(np.array([scores[i] for i in range(graph.population)], dtype=int), mode=normalization_mode)
 
 
-def stationary_probabilities(graph: Graph, normalize: str = 'gaussian') -> np.array:
+def stationary_probabilities(graph: Graph, normalization_mode: str = 'gaussian') -> np.array:
```
thoward27, line 194: I think we should use this: https://docs.python.org/3/library/typing.html#typing.Literal

pyclam/tests/utils.py
```diff
@@ -15,24 +15,3 @@ def linear_search(point: Data, radius: Radius, data: Data, metric: str):
         distances = cdist(point, batch, metric)[0]
         results.extend([p for p, d in zip(batch, distances) if d <= radius])
     return results
-
-
-def trace_lineage(cluster: Cluster, other: Cluster):  # TODO: Cover
```
thoward27, line 20: Is this still functional? Do we see a potential need for it in the future?

pyclam/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+def catch_normalization_mode(mode: str) -> None:
+    modes: List[str] = ['linear', 'gaussian', 'sigmoid']
+    if mode not in modes:
+        raise ValueError(f'Normalization method {mode} is undefined. Must by one of {modes}.')
+    else:
+        return
+
+
+def normalize(scores: np.array, mode: str) -> np.array:
```
thoward27, line 18: docstring

pyclam/utils.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
+
+import numpy as np
+from scipy.special import erf
+
+SUBSAMPLE_LIMIT = 100
+BATCH_SIZE = 10_000
+
+
+def catch_normalization_mode(mode: str) -> None:
```
thoward27, line 10: docstring

pyclam/utils.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
```
thoward27, line 1: maybe add module docstring

pyclam/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+import numpy as np
+from scipy.special import erf
+
+SUBSAMPLE_LIMIT = 100
+BATCH_SIZE = 10_000
```
thoward27, line 7: I don't see a direct use of these in this file. Perhaps we should consider a `constants.py` file

thoward27 (changes requested):
requesting changes for docstrings, also had a typing nit and a suggestion around constants

pyclam/chaoda.py
nishaq503: I think we should have a separate PR for using literals. I think we should treat this more like we treat Metric. some strings to represent build-in functionality, while also allowing a user-defined function. In either case, I think that's outside the scope of this PR

pyclam/tests/utils.py
nishaq503: It's not used anywhere and I don't see it being used anywhere in the future. If I'm wrong, it's still in the git history

pyclam/utils.py
nishaq503: We can talk about a constants.py or a config.py file but I think that such a thing should be its own PR

nishaq503 (commented):
I think that using the Literal type for normalization modes, and maybe also fro Metric, is a good idea. Also, breaking out constants and some other things, like setting up logging, into a config.py file will be a good idea.

However, I think that those changes should be their own PR and are ouside the scope of this PR.

thoward27 (approved):
LGTM

PR #38, opened 2020-11-12
Cluster ratios are now in Manifold

nishaq503:
because we can now properly normalize the ratios and make them dataset-independent.

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -11,7 +11,7 @@ import numpy as np
 from scipy.spatial.distance import cdist
 
 from pyclam.types import Data, Radius, Vector, Metric
-from pyclam.utils import BATCH_SIZE, SUBSAMPLE_LIMIT
+from pyclam.utils import BATCH_SIZE, SUBSAMPLE_LIMIT, EPSILON, normalize
 
 LOG_LEVEL = logging.INFO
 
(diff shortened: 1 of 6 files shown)


pyclam/manifold.py
```diff
@@ -849,6 +822,46 @@ def clear_cache(self):
...
 
+    @property
+    def ordered_clusters(self) -> List[Cluster]:
+        if 'ordered_clusters' not in self.cache:
+            self.cache['ordered_clusters'] = [
+                cluster for layer in self.layers
+                for cluster in layer.clusters
+                if cluster.depth == layer.depth
+            ]
+        return self.cache['ordered_clusters']
```
thoward27, line 833: docstring

pyclam/manifold.py
```diff
@@ -849,6 +822,46 @@ def clear_cache(self):
...
+
+    @property
+    def ratios(self) -> Dict[Cluster, np.array]:
```
thoward27, line 836: docstring

pyclam/manifold.py
```diff
@@ -849,6 +822,46 @@ def clear_cache(self):
...
+            for i in range(cluster_ratios.shape[1]):
+                cluster_ratios[:, i] = normalize(cluster_ratios[:, i], 'gaussian')
+
+            self.cache['ratios'] = {
+                cluster: cluster_ratios[cluster_indices[cluster]]
+                for cluster in self.ordered_clusters
+            }
+        return self.cache['ratios']
+
+    def cluster_ratios(self, cluster: Cluster) -> np.array:
```
thoward27, line 862: docstring

pyclam/manifold.py
```diff
@@ -849,6 +822,46 @@ def clear_cache(self):
...
+                        cluster.cardinality / cluster.parent.cardinality,
+                        cluster.radius / (cluster.parent.radius + EPSILON),
+                        cluster.local_fractal_dimension / cluster.parent.local_fractal_dimension,
+                    ], dtype=float)
+                    previous = cluster_ratios[cluster_indices[cluster.parent]][3:]
+                    new = alpha * current + (1 - alpha) * previous
+                    cluster_ratios[cluster_indices[cluster]] = np.concatenate([current, new])
+
+            for i in range(cluster_ratios.shape[1]):
+                cluster_ratios[:, i] = normalize(cluster_ratios[:, i], 'gaussian')
```
thoward27, line 854: do we *always* want to normalize eagerly? should this be a toggle?

thoward27 (approved):
Needs docstrings, LGTM otherwise. Surprised by how much I like this change. Some of these rations span clusters *already* making them great candidates for manifold already. Top that off with the ability to normalize across the board- rad.

pyclam/manifold.py
nishaq503: done

pyclam/manifold.py
nishaq503: done

pyclam/manifold.py
nishaq503: done

pyclam/manifold.py
nishaq503: chaoda performance was always worse when the ratios were not normalized. I'm open for adding a toggle for normalization but it would complicate caching. Once we figure out that complexity, we can add it

PR #39, opened 2020-11-12
Dot-File functionality for Graph

nishaq503:
These methods will later be used to write dump/load methods for Graph.

feat: as-dot-string and from-dot-string methods for Graph

### DIFF
--- a/pyclam/manifold.py
+++ b/pyclam/manifold.py
@@ -5,7 +5,7 @@ import logging
 import pickle
 from collections import deque
 from operator import itemgetter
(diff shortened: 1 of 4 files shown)


pyclam/manifold.py
```diff
@@ -569,6 +567,96 @@ def clear_cache(self) -> None:
         self.cache.clear()
         return
 
+    def as_dot_string(
```
thoward27, line 570: I don't know what I'm looking for here, this function looks OK, but a large part of that is just trusting you

thoward27 (approved):
LGTM with the caveat that I don't know what a correct dot string *would* be.

PR #40, opened 2020-11-13
Added Labels as a Selection Criterion

nishaq503:
Selects clusters that have an overwhelming majority of points from one class.

feat: label based selection criteria

### DIFF
--- a/pyclam/criterion.py
+++ b/pyclam/criterion.py
@@ -1,11 +1,12 @@
 import logging
 from abc import ABC, abstractmethod
-from typing import Set, Tuple, List, Union, Dict, Callable
+from typing import Set, Tuple, List, Union, Dict, Callable, Any
 
 import numpy as np
 from scipy.spatial.distance import cdist
 
 from pyclam.manifold import Cluster, Manifold
+from collections import Counter
 
 
 class Criterion(ABC):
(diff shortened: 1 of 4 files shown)


pyclam/criterion.py
```diff
@@ -337,3 +337,36 @@ def _select(root: Cluster, upper: float, lower: float, grace_depth: int) -> Set[
...
+
+
+class Labels(SelectionCriterion):
+    def __init__(self, labels: Dict[int, int]):
+        """ Uses a given dictionary of index -> label to select optimal clusters.
+
+        During a BFT of the tree, when a cluster has an overwhelming majority of points with the same label,
+        that cluster is selected.
+        If some data points are unlabelled, then when a cluster contains no labelled points, it too is selected.
+        """
```
thoward27, line 349: We don't usually document in `__init__`, usually docstrings go directly on the class object

pyclam/criterion.py
```diff
@@ -337,3 +337,36 @@ def _select(root: Cluster, upper: float, lower: float, grace_depth: int) -> Set[
             inactive = {child for cluster in inactive for child in cluster.children}
 
         return selected
+
+
+class Labels(SelectionCriterion):
+    def __init__(self, labels: Dict[int, int]):
+        """ Uses a given dictionary of index -> label to select optimal clusters.
+
+        During a BFT of the tree, when a cluster has an overwhelming majority of points with the same label,
```
thoward27, line 346: What is an overwhelming majority? Can we parameterize that?

pyclam/criterion.py
```diff
@@ -337,3 +337,36 @@ def _select(root: Cluster, upper: float, lower: float, grace_depth: int) -> Set[
...
+            else:
+                labels: List[int] = [self.labels[p] for p in cluster.argpoints if p in self.labels]
+                if len(labels) > 0:
+                    max_fraction = max(dict(Counter(labels)).values()) / cluster.cardinality
+                    if max_fraction > self.threshold:
+                        selected.add(cluster)
+                    else:
+                        frontier.update(cluster.children)
+                else:
+                    # TODO: Is this really the best idea?
```
thoward27, line 370: Seems like the only good path. We can't expect that future children will suddenly have more labels. Unless you just take these types of situations down to singletons? What other ideas did you have?

thoward27 (changes requested):
I'd like to see the overwhelming majority bit better defined in our documentation and potentially parameterized if possible.

pyclam/criterion.py
nishaq503: moved

pyclam/criterion.py
nishaq503: It was already parameterized in self.threshold. I added an explanation to the doc-string.

pyclam/criterion.py
nishaq503: reverting to other selection criteria?

pyclam/criterion.py
```diff
@@ -392,3 +393,40 @@ def __call__(self, root: Cluster) -> Set[Cluster]:
...
+class Labels(SelectionCriterion):
+    """ Uses a given dictionary of index -> label to select optimal clusters.
+
+    During a BFT of the tree, when a cluster has an overwhelming majority of points with the same label,
+    that cluster is selected.
+    If some data points are unlabelled, then when a cluster contains no labelled points, it too is selected.
+
+    The 'overwhelming majority' fraction is parameterized by the fraction of the minority class.
+    Let f be the fraction of the smallest minority class in all the data.
+    Then the threshold for determining an overwhelming majority is (1 - f / 10).
```
thoward27, line 407: I meant you should parameterize the magical `10`, unless there's a good reason for it to be 10?

thoward27 (commented):
I don't think our definitions of parameterized quite lined up, let me know your thoughts

pyclam/criterion.py
nishaq503: 10 is a very good number here so I am going to choose an appropriate default value for the argument

thoward27 (approved):
LGTM 🎉

PR #42, opened 2020-11-14
Jaccard Index methods for Cluster and Graph

nishaq503:
The Jaccard index is going to be useful for thinking a bit differently about persistent components.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "clam"
-version = "0.3.2"
+version = "0.3.3"
 authors = ["Tom Howard <info@tomhoward.codes>", "Najib Ishaq <nishaq@my.uri.edu>", "Noah Daniels <noah_daniels@uri.edu>"]
 edition = "2018"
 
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "clam"
-version = "0.3.2"
+version = "0.3.5"
```
thoward27, line 3: make sure you address this before merging it in. Whatever goes first should be `0.3.3`

pyclam/manifold.py
```diff
@@ -105,6 +105,12 @@ def __contains__(self, point: Data) -> bool:
         """ Check weather the given point could be inside this cluster. """
         return self.overlaps(point=point, radius=0.)
 
+    def jaccard(self, other: 'Cluster') -> float:
+        """ The Jaccard Index between two Clusters. """
```
thoward27, line 109: I think a little text about what the jaccard index is would be nice to provide, but certainly not necessary.

thoward27 (approved):
LGTM! Another really easy one.

PR #43, opened 2020-11-17
Added tree-search-history method

nishaq503:
This method, in addition to returning the hits for regular tree-search, also returns the history of all candidate clusters at each depth during the search.

This is going to be useful for supervised learning in CHAODA.


thoward27 (approved):
LGTM

PR #45, opened 2020-11-20
Created new locks for poetry

nishaq503:
because numpy 1.19 contains an API-breaking change that breaks tensorflow


nishaq503:
will reopen after other PR is done

PR #46, opened 2020-11-20
locking numpy under 1.19

nishaq503:
1.19 contains an API breaking change that causes problems with tensorflow. I see CLAM being used in conjunction with TF quite often.


thoward27 (approved):
LGTM

PR #49, opened 2021-01-03
Dataset choose_unique

thoward27:


src/dataset.rs
```diff
@@ -56,36 +59,64 @@ impl<T: Number, U: Number> Dataset<T, U> {
...
+    /// Returns `n` unique random indices from the provided vector.
+    /// If `n` is greater than the number of indices provided, the full list is returned in shuffled order.
     #[allow(clippy::ptr_arg)]
     pub fn choose_unique(&self, indices: Indices, n: usize) -> Indices {
         // TODO: actually check for uniqueness among choices
-        indices.into_iter().choose_multiple(&mut rand::thread_rng(), n)
+        let mut x = indices;
+        x.shuffle(&mut rand::thread_rng());
+        x.truncate(n);
+        x
```
nishaq503, line 96: This is functionally identical to the old version. choose_multiple already chooses without replacement. We need to make certain that no pair of indices correspond to points whose distance apart is zero.

nishaq503 (changes requested):
The cleanup and docs look really good. I've been holding off on the docs because the API has been very fluid so far, but I think it's time to start adding proper docs.

Still, the choose_unique method needs to choose unique points instead of unique indices.

PR #58, opened 2021-03-02
feat: dataset as a trait

nishaq503:
Implemented dataset as a trait.
Also ran rustfmt


thoward27 (approved):
Little bit of a noisy diff with the amount of code that got formatted, but I don't see anything wrong with this merging to dev. LGTM!

PR #59, opened 2021-03-04
feat: Fasta struct

nishaq503:
This struct adds a struct that reads a memory-mapped Fasta file and only loads instances into memory when needed.

The tests pass locally but are commented out because I can't find a small enough fasta file to upload with the repo

### DIFF
--- a/src/sample_datasets.rs
+++ b/src/sample_datasets.rs
@@ -1,0 +1,371 @@
+use std::{fmt, result};
+
+use dashmap::DashMap;
+use ndarray::{Array2, ArrayView, Axis, IxDyn};
+use rand::seq::IteratorRandom;
+use rayon::prelude::*;
+
+use crate::dataset::Dataset;
+use crate::metric::{Metric, metric_new, Number};
+use crate::types::{Index, Indices};
+use std::path::PathBuf;
+use std::fs::File;
+use memmap2::Mmap;
+
+/// RowMajor dataset represented as a 2-dimensional array where rows are instances and columns are attributes.
+///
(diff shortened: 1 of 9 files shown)


src/sample_datasets.rs
```diff
@@ -0,0 +1,371 @@
...
+        metric: &'static str,
+        path: PathBuf,
+        shape: &'static [usize],
+        offset: usize,
+    ) -> Result<Self, String> {
+        let file = match File::open(path) {
+            Ok(f) => f,
+            Err(_) => return Err("Could not open file".to_string()),
+        };
+        let mmap = unsafe { match Mmap::map(&file) {
```
nishaq503, line 194: unsafe because the file being read requires a file-system lock. Basically, don't change the file out from under this reader.

src/sample_datasets.rs
```diff
@@ -0,0 +1,371 @@
...
+            num_sequences: shape[0],
+            seq_len: shape[1],
+            offset,
+            function: metric_new(metric)?,
+            cache: DashMap::new(),
+        })
+    }
+}
+
+impl Dataset<u8, u64> for Fasta {
```
nishaq503, line 212: Some more of these methods could be defaults in the trait but I need to implement this for a few other structs before I have a decent idea for how to get that done

thoward27 (commented):
Curious why we're dropping the use of arc?

nishaq503:
Because you can't return an Arc to a piece of a MMap and also because you can't return a reference to a local variable from a function. Both of those reasons became relevant in the implementation of the instance method.

PR #66, opened 2021-05-03
review: changes after first round of review for paper

nishaq503:
building graphs in .fit method instead.
fixed transition matrix.
removed random-walks and added vertex-degree.
cleaner meta-ml.select. method
faster edges-dict by iterating only once on edges.
updated dependency versions
added class variable of method names
using a criterion-function for each pair of metric/method
fix: now using method-graph pairs as well
fix: criterion.Layer usage in Search

### DIFF
--- a/pyclam/criterion.py
+++ b/pyclam/criterion.py
@@ -144,19 +144,25 @@ class MetaMLSelect(SelectionCriterion):
 
         :param predict_auc: A function that takes a cluster's ratios and returns a predicted auc for selecting that cluster.
         """
-        self.predict_auc: Callable[[np.array], float] = predict_auc
         if min_depth < 1:
             raise ValueError(f'min-depth must be a positive integer.')
         self.min_depth = min_depth
 
-    def __call__(self, root: Cluster) -> Set[Cluster]:
-        def predict_auc(cluster):
-            # adjust predicted auc by cardinality to slightly favor clusters with high cardinalities.
-            return self.predict_auc(manifold.cluster_ratios(cluster)) * (1 + cluster.cardinality / root.cardinality / 10)
+        # TODO: Investigate discount factors to favor high-cardinality clusters
+        #  measure performance change vs runtime improvement for individual methods in CHAODA
+        # def discount_factor(cluster: Cluster):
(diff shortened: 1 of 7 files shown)


pyclam/criterion.py
```diff
@@ -144,19 +144,25 @@ def __init__(self, predict_auc: Callable[[np.array], float], min_depth: int = 4)
...
-            # adjust predicted auc by cardinality to slightly favor clusters with high cardinalities.
-            return self.predict_auc(manifold.cluster_ratios(cluster)) * (1 + cluster.cardinality / root.cardinality / 10)
+        # TODO: Investigate discount factors to favor high-cardinality clusters
+        #  measure performance change vs runtime improvement for individual methods in CHAODA
+        # def discount_factor(cluster: Cluster):
+        #     num_points = cluster.manifold.root.cardinality
+        #     return 1 + ((cluster.cardinality / num_points) / np.sqrt(num_points))
+
+        def ratios(cluster: Cluster):
+            return cluster.manifold.cluster_ratios(cluster)
```
thoward27, line 158: this is some crazy redirection loll

pyclam/manifold.py
```diff
@@ -564,8 +570,8 @@ def as_matrix(self) -> Tuple[List[Cluster], np.array]:
         matrix: np.array = np.zeros(shape=(len(clusters), len(clusters)), dtype=float)
         for edge in self.edges:
             i, j = indices[edge.left], indices[edge.right]
-            matrix[i][j] = edge.distance
-            matrix[j][i] = edge.distance
+            matrix[i][j] = 1. / edge.distance
+            matrix[j][i] = 1. / edge.distance
```
thoward27, line 574: Interesting diff, intentional?

thoward27 (approved):
LGTM. Some weird diffs in this one, not sure I entirely understand each change, but the overall delta appears to be functional

pyclam/criterion.py
nishaq503: I know right

pyclam/manifold.py
nishaq503: Yes. This was a bug

PR #67, opened 2021-05-03
added folder modules and prelude

nishaq503:
added module for cakes
added module for chaoda
using prelude
reduced reliance on DashMap


src/cakes/search.rs
```diff
@@ -170,8 +166,8 @@ impl<T: Number, U: Number> Search<T, U> {
 mod tests {
     use std::sync::Arc;
 
-    use crate::dataset::Dataset;
-    use crate::dataset::RowMajor;
+    use crate::clam::Dataset;
+    use crate::clam::dataset::RowMajor;
```
thoward27, line 170: I'm curious why dataset is falling into the clam mod

src/clam/graph.rs
```diff
@@ -120,7 +121,7 @@ impl<T: Number, U: Number> Graph<T, U> {
             indices: vec![],
             depth: 0,
             min_depth: 0,
-            edges_dict: Arc::new(DashMap::new()),
+            edges_dict: Arc::new(HashMap::new()),
```
thoward27, line 124: Any reasoning for this transition?

thoward27 (approved):
LGTM. But I will say I'm not entirely sold on having core be within a nested module. I see why you've made the decision, but still lean towards having "core" outside of a nested mod context

src/clam/graph.rs
nishaq503: DashMaps are mutable by default. I want to expose these as HashMap/HashSet while internally working with DashMap/DashSet for concurrent processing

nishaq503: This is the first step to that goal. I think I may have figured out compression but it requires a lot of changes to Cluster so I'm going to finish this transition after I'm done with compression

src/cakes/search.rs
nishaq503: Dataset is part of the core framework though I could be persuaded to treat Dataset (and Metric) differently. Dataset and metric undergird Cluster, Graph, and Manifold

nishaq503 (commented):
It was getting too cluttered in src for me. I could be persuaded to move clam back out although it's about to get a new member when compression is done

thoward27 (approved):
LGETM, not totally sold on search -> cakes, but 🤷  we can always rename things

PR #69, opened 2021-05-12
working on compression

nishaq503:
Successfully created an encodable dataset with the Hamming distance.
Created modules for traits and core.
Metric is not a trait to allow for encode/decode methods.
more documentation.
cleaner benchmarking messages.


thoward27:
@nishaq503 what's the status of this PR? Do you still need my review here?

thoward27:
@nishaq503 re-request my review if you still want it here :) just need to keep my TODO list clean

PR #75, opened 2022-03-10
applied lots of fixes

nishaq503:
Debugged algorithms and applied lots of fixes. ROC scores are now on par with python implementation. There are still some discrepancies in speed for various algorithms. Some are much faster than the python versions but others are slower. This is because of how the Graph class/struct is treaded in python/rust. I have not yet spent much time optimizing the Rust version as I did for Python. I will make performance improvements later.


debug.py
```diff
@@ -0,0 +1,33 @@
+import pathlib
```
thoward27, line 1: Maybe nix this file and gitignore it

thoward27 (approved):
LGTM, I could not spot any glaring weaknesses here, though I definitely spotted a lot of inspect calls and other odds and ends to get cleaned up later

PR #77, opened 2022-04-04
propagated min-depth param for MetaMLSelect to Chaoda

nishaq503:
This is necessary for when CHAODA always selects clusters that create a Graph with only one connected component.


thoward27 (approved):
LGTM

PR #80, opened 2022-05-15
Implemented a supervised classifier

nishaq503:
This PR implements a supervised classifier using CHAODA and Search.
This required for me to add methods to predict anomaly scores for new data in CHAODA.
The method, using the meta-ml models trained for anomaly detection, has an accuracy of 0.914 on the digit-recognition dataset in sklearn.
I am yet to add a method for training new meta-ml methods for classification.


pyclam/search.py
```diff
@@ -136,10 +144,19 @@ def tree_search(self, query: Datum, radius: float, start: Cluster, *, max_depth:
...
+    def tree_search_history(
+            self,
+            query: Datum,
+            radius: float,
+            *,
+            start: Cluster = None,
+            max_depth: int = None,
+    ) -> Tuple[ClusterResults, ClusterResults]:
         """ Same as tree-search, except that it also returns the history of candidate clusters at each depth. """
+        start = self.root if start is None else start
```
thoward27, line 158: could do `start = start or self.root` here

pyclam/search.py
```diff
@@ -122,7 +130,7 @@ def _parse_query(self, query: Datum) -> np.ndarray:
         self._check_shape(query)
         return query
 
-    def tree_search(self, query: Datum, radius: float, start: Cluster, *, max_depth: Optional[int] = None) -> ClusterResults:
+    def tree_search(self, query: Datum, radius: float, *, start: Cluster = None, max_depth: int = None) -> ClusterResults:
```
thoward27, line 133: Hmm, I see you've trimmed a lot of `Optional`'s off of places that would happily except `None`, which seems very optional IMO. I would be very curious why you've made these changes!

thoward27 (approved):
I didn't catch anything functionally wrong, noticed a few nits. 

Overall, this seems really clever, using separating the labels into distinct manifolds then computing anomalousness scores on the manifolds ("bowls") to figure out what manifold a point is most similar to. Neat!

pyclam/search.py
nishaq503: Done.

pyclam/search.py
nishaq503: Giving a type-hint and then setting the default value to None means that at interpretation time, that time automatically gets an Optional put around it.

pyclam/classifier.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, **kwargs):
+        """ Creates and initializes a CLAM Classifier.
+
+        See CHAODA for the list of input arguments.
+        """
+        self.kwargs = kwargs
+
+        self._bowls: typing.Dict[int, chaoda.CHAODA] = dict()
+
+    def fit(self, data: numpy.ndarray, labels: typing.List[int], *, voting: str = 'mean') -> 'Classifier':
```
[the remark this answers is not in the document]
nishaq503: Literals need to be used in a lot of places in the code. It'd be better to have a separate PR for that.

ndaniels:
Wow, the modern Python type system is just voodoo. I don’t like it.

pyclam/chaoda.py
```diff
@@ -284,45 +301,41 @@ def vertex_degree(self, graph: Graph) -> Scores:
...
 
     # noinspection PyMethodMayBeStatic
-    def _vote(self, scores: np.array, mode: str):
+    def _vote(self, scores: np.array):
         """ Vote among all individual-scores for ensemble-score.
 
         :param scores: An array of shape (num_graphs, num_points) of scores to be voted among.
-        :param mode: Voting mode to use. Must be one of VOTING_MODES
         """
-        if mode not in VOTING_MODES:
```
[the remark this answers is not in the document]
nishaq503: This is in a hidden method so I don't feel strongly about it either way. The check is performed in the `fit` method. I've added a `NotImplementedError` in the else block here.

PR #83, opened 2022-05-23
Updated Logging and Removed Poetry

nishaq503:
The log-level for the package can now be set via an environment variable called `CLAM_LOG`.
Each important file where we need to log something has its own named logger object.

The logger objects make it so that a user can grab the individual objects and reconfigure logging. For example, someone can now redirect logs to a file on disk or to some logs tracking package for remote deployment.

This PR also completely removes the files related to poetry because we switched to using `setup.py` in an earlier PR.
I have also added stricter version bounds for required packages in `setup.py`.


.github/workflows/tests.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - name: Dependencies
         run: |
           python -m pip install --upgrade pip setuptools wheel
-          pip install -r requirements.txt
+          pip install twine toml
```
thoward27, line 26: don't need twine

pyclam/anomaly_detection/chaoda.py
```diff
@@ -5,15 +5,23 @@
...
 from ..core import types
+from ..search import CAKES
+from ..utils import constants
 from ..utils import helpers
 
+logging.basicConfig(
+    format="%(asctime)s - %(levelname)s - %(name)s - %(module)s.%(funcName)s - %(message)s",
+    datefmt='%d-%b-%y %H:%M:%S',
+)
+logger = logging.getLogger('chaoda')
```
thoward27, line 22: should use relpath or some other non-human-managed key

pyclam/classification/classifier.py
```diff
@@ -4,6 +4,14 @@
 import numpy
 
 from ..anomaly_detection import CHAODA
+from ..utils import constants
+
+logging.basicConfig(
+    format="%(asctime)s - %(levelname)s - %(name)s - %(module)s.%(funcName)s - %(message)s",
```
thoward27, line 10: Should probably pull the basic config out into constants/settings

pyclam/classification/classifier.py
```diff
@@ -4,6 +4,14 @@
...
+    datefmt='%d-%b-%y %H:%M:%S',
```
thoward27, line 11: this string too

thoward27 (approved):

PR #94, opened 2022-09-08
Rust metaml

ijchen:
Code is ready for review. Known issues:
- [x] Line endings for at least metaml.rs may be CRLF instead of LF
- [x] Visibility in mod.rs was not considered thoughtfully
- [x] Incorporate initial PR feedback
- [x] Refactor `MetaMLDataset` to avoid relying on ndarray except in `from_npy(...)`
- [x] Use `[T; 6]`s where possible
- [x] Get a better target name from Najib than Score
- [x] Update doctests to read from disk where reasonable
- [x] Document all `no_run`s with reasoning
- [x] Resolve copy discussion https://github.com/URI-ABD/clam/pull/94#discussion_r1018545016
- [x] Figure out why this branch is showing as "cannot be rebased"
- [x] Figure out why CI is not running


src/anomaly/metaml.rs
```diff
@@ -0,0 +1,214 @@
...
+///
+/// // Do something with the dataset
+/// todo!();
+/// ```
+pub fn create_dataset_from_npy(features_file_path: &str, targets_file_path: &str) -> MetaMLDataset {
+    // Read in the training data
+    let features_f64: Array2<f64> = read_npy(features_file_path).expect("failed to read the features data file");
+    let targets_f64: Array1<f64> = read_npy(targets_file_path).expect("failed to read the outputs data file");
+
+    // Transform the training data from f64 to f32 (we are given f64, but automl uses f32s)
```
thoward27, line 79: This is the second part of the two-part advice... this comment is **very** useful.

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+    SupervisedModel,
+};
+use smartcore::dataset::Dataset;
+
+use super::metaml::{MetaMLDataset, MetaMLModel};
+
+// Transforms a MetaMLDataset into an automl dataset
+fn metaml_dataset_to_automl(data: MetaMLDataset) -> Dataset<f32, f32> {
+    // Double-check meta-information about the data
+    assert!(data.features.ncols() == 6);
```
thoward27, line 14: Seeing this again, I feel that the PR would benefit from a function that handled these assertions. Something like `assert_correct_shape`. That way, we could aggregate the TODO for compile-time shape checking into one place. Would make it easier to grep for the places we need to update.

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+    // Data labels
+    let feature_names = vec![
+        "Cardinality".to_string(),
+        "Radius".to_string(),
+        "Local fractal dimension".to_string(),
+        "Cardinality (exponential moving average)".to_string(),
+        "Radius (exponential moving average)".to_string(),
+        "Local fractal dimension (exponential moving average)".to_string(),
+    ];
+    let target_names = vec!["Score".to_string()];
```
thoward27, line 26: Tell Najib to come up with a better target name.

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+/// A metaml wrapper for an [automl] linear regressor
+pub struct LinearRegressor {
+    model: Option<SupervisedModel>,
+}
+
+impl LinearRegressor {
+    /// Constructs a new LinearRegressor
+    ///
+    /// # Examples
+    /// ```no_run
```
thoward27, line 51: See my comments in `metaml.rs` on `no_run` and `todo!`

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// # use clam::anomaly::automl_regressors::LinearRegressor;
+    /// #
+    /// // Construct a new LinearRegressor
+    /// let regressor = LinearRegressor::new();
+    ///
+    /// // Train and use your regressor
+    /// todo!();
```
thoward27, line 58: See my comments in `metaml.rs` on `no_run` and `todo!`

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+/// A metaml wrapper for an [automl] decision tree regressor
+pub struct DecisionTreeRegressor {
+    model: Option<SupervisedModel>,
+}
+
+impl DecisionTreeRegressor {
+    /// Constructs a new DecisionTreeRegressor
+    ///
+    /// # Examples
+    /// ```no_run
```
thoward27, line 111: See my comments in `metaml.rs` on `no_run` and `todo!`

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// # use clam::anomaly::automl_regressors::DecisionTreeRegressor;
+    /// #
+    /// // Construct a new DecisionTreeRegressor
+    /// let regressor = DecisionTreeRegressor::new();
+    ///
+    /// // Train and use your regressor
+    /// todo!();
```
thoward27, line 118: See my comments in `metaml.rs` on `no_run` and `todo!`

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[derive(Default)]
+/// A metaml wrapper for an [automl] linear regressor
+pub struct LinearRegressor {
+    model: Option<SupervisedModel>,
+}
+
+impl LinearRegressor {
+    /// Constructs a new LinearRegressor
```
thoward27, line 48: Instead of documenting the `new` method, move this documentation to the struct itself.

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+
+#[derive(Default)]
+/// A metaml wrapper for an [automl] decision tree regressor
+pub struct DecisionTreeRegressor {
+    model: Option<SupervisedModel>,
+}
+
+impl DecisionTreeRegressor {
+    /// Constructs a new DecisionTreeRegressor
+    ///
```
thoward27, line 109: Instead of documenting the `new` method, move this documentation to the struct itself.

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl MetaMLModel for DecisionTreeRegressor {
+    fn train(&mut self, data: MetaMLDataset) {
+        // Construct a metaml dataset from the input data
+        let dataset = metaml_dataset_to_automl(data);
+
+        // Create the settings and model
+        let settings = automl::Settings::default_regression()
+            .only(Algorithm::DecisionTreeRegressor)
+            .with_decision_tree_regressor_settings(DecisionTreeRegressorParameters::default().with_max_depth(3));
```
thoward27, line 133: We should make `max_depth` a member of the `DecisionTreeRegressor` struct rather than just hard-coding it here

src/anomaly/automl_regressors.rs
```diff
@@ -0,0 +1,161 @@
...
+    fn load(path: &str) -> Self {
+        let model = SupervisedModel::new_from_file(path);
+        DecisionTreeRegressor { model: Some(model) }
+    }
+
+    fn save(&self, path: &str) {
+        let model = self.model.as_ref().expect("Model must be trained before being saved");
+        model.save(path);
+    }
+}
```
thoward27, line 161: Diff view shows no trailing newline here

thoward27 (changes requested):
Overall this PR looks pretty good, great job on the documentation for the code. I had some feedback around types in `metaml` and some suggestions for your code blocks in docstrings. 

PS my condolences for having to work with Najib...

ijchen:
@thoward27 @nishaq503 I have made revisions based on your feedback, and I believe the code is ready for re-review. Let me know what you think :)

src/anomaly/automl_regressors.rs
```diff
@@ -23,18 +26,18 @@ fn metaml_dataset_to_automl(data: MetaMLDataset) -> Dataset<f32, f32> {
...
     Dataset {
-        num_samples: data.features.nrows(),
+        num_samples: data.features.len(),
         num_features: 6,
         feature_names,
         target_names,
         description,
-        data: data.features.into_raw_vec(),
-        target: data.targets.into_raw_vec(),
+        data: data.features.into_iter().flatten().collect(),
```
thoward27, line 39: This is a data copy, right? If so, we do not want data copies.

src/anomaly/metaml.rs
```diff
@@ -1,214 +1,267 @@
...
+    fn predict(&self, features: &[f32; 6]) -> f32;
+
+    /// Loads a trained meta-ml model from disk.
+    ///
+    /// # Panics
+    /// * If the serialized model cannot be read from the input file path
+    /// * If the trained model cannot be deserialized
+    ///
+    /// # Examples
+    /// ```no_run
```
thoward27, line 97: Nix the no_run or add a comment about why this cannot be run, ie "I cannot get x to work because of y, so this is disabled for now"

src/anomaly/metaml.rs
```diff
@@ -1,214 +1,267 @@
...
+
+    /// Saves a trained meta-ml model to disk
+    ///
+    /// # Panics
+    /// * If the model hasn't been trained
+    /// * If the trained model cannot be serialized
+    /// * If the serialized model cannot be written to the output file path
+    ///
+    /// # Examples
+    /// ```no_run
```
thoward27, line 128: see above

src/anomaly/metaml.rs
```diff
@@ -1,214 +1,267 @@
...
+    /// * If the number of rows in the features data doesn't match the number of
+    /// elements in the targets data
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # use clam::anomaly::metaml::MetaMLDataset;
+    /// # use std::path::Path;
+    /// #
+    /// // File paths to training data in the numpy .npy file format
+    /// let features_path = Path::new("path/to/data/features.npy");
```
thoward27, line 215: Can we add some data to this repo so that snippets like this could be run? See the data folder for examples

thoward27 (commented):
Looking good, few more minor changes/comments for you

thoward27:
@ijchen work with @nishaq503 (he volunteered) to figure out why this branch is showing as "cannot be rebased" and why CI is not running (the ci part is most likely @nishaq503's fault anyways).

src/anomaly/automl_regressors.rs
ijchen: **TL;DR** This isn't doing any more data copying than before (just more explicitly), and optimizing to avoid any data copying would be a pretty big hit to code coupling (but I'll do it if you want me to anyway of course :))

At a high level, no, this is not a data copy. `.into_iter()`, `.flatten()`, and `.collect()` all work with owned values here.

At a more low level, this *might* be doing some data copying in `.collect()` due to the `.flatten()` (if the underlying data wasn't previously contiguous in memory). If this were the case though, that same copying would have been occurring elsewhere - just less explicitly since it's not in our code, but in another function we call somewhere. This copying is necessary to exist *somewhere* because we're working with two different APIs (`automl` and `ndarray-npy`) which give and expect the same data in different formats. This change does the data format transformation when creating the MetaMLDataset instead of when using it.

If this is a performance issue, we could modify our code so that MetaMLDataset's internal representation is tied to the output format from `ndarray-npy`, and modify `automl`'s code to expect that same format. This will eliminate the need for copying, but will significantly increase the amount of coupling and cost of changes/expansion in the future... in my opinion, that tradeoff isn't worth it unless we find through benchmarking that this is actually a performance issue. If you feel differently, absolutely let me know and I'll refactor the code with performance in mind (although that will take some time)

ndaniels: This is the correct answer: profile and benchmark, and see if there's actually any performance bottleneck worth dealing with. I somehow doubt it.

ijchen:
@thoward27 I just got out of a meeting with Najib to fix the rebasing and CI issues, and he was able to solve both problems. He and I both believe the code is now ready for re-review. Specific issues he fixed:
- Najib fixed the issue for rebasing
- CI was not working because this was my fork, where actions had not yet been enabled. This has been fixed

thoward27 (approved):

PR #98, opened 2023-03-16
Implemented reordering logic into new dataset schema

olwmc:
*Assigned to me by Najib*

## Overview
This PR implements dataset reordering logic for `VecVec`. A new function in the `Dataset` trait called `reorder` takes a list of indices and applies that list of indices to the dataset like a permutation.

## Tests
`core::dataset::test_reordering` - Tests table reordering on a randomly generated three-dimensional dataset with increasing lengths. The maximum length could be decreased before merging to speed up test time. Currently running this test takes ~0.25 seconds on my laptop.

## Questions
- [ ] Should I keep the documentation? I'm not sure what the documentation style of this project is.
- [ ] Should indices be validated? I.e. we would reject index lists that contained duplicates, numbers out of range (n > cardinality), etc.
- [ ] Should this be implemented in the `Dataset` trait or just `VecVec`? I imagine we would like to do table reordering on all datasets, so I implemented it for `Dataset`, though I'm not sure.

### DIFF
--- a/src/core/dataset.rs
+++ b/src/core/dataset.rs
@@ -12,6 +12,16 @@ pub trait Dataset<T: Number, U: Number>: std::fmt::Debug + Send + Sync {
     fn one_to_one(&self, left: usize, right: usize) -> U;
     fn query_to_one(&self, query: &[T], index: usize) -> U;
 
+    /// Swaps the values at two given indices in the dataset.
+    ///
+    /// Note: It is acceptable for this function to panic if `i` or `j` are not valid indices in the
+    /// dataset.
+    ///
+    /// # Arguments
+    /// `i` - An index in the dataset
+    /// `j` - An index in the dataset
+    fn swap(&mut self, i: usize, j: usize);
+
     fn are_instances_equal(&self, left: usize, right: usize) -> bool {
         self.one_to_one(left, right) == U::zero()
     }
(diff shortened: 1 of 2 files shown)


src/core/dataset.rs
```diff
@@ -69,6 +69,12 @@ pub trait Dataset<T: Number, U: Number>: std::fmt::Debug + Send + Sync {
 
         chosen
     }
+
+    /// Reorders the internal dataset by a given permutation of indices
+    ///
+    /// # Arguments
+    /// `indices` - A permutation of indices that will be applied to the dataset
+    fn reorder(&mut self, indices: &Vec<usize>);
```
nishaq503, line 77: The type of `indices` should be `&[usize]`

src/core/dataset.rs
```diff
@@ -125,13 +131,49 @@ impl<T: Number, U: Number> Dataset<T, U> for VecVec<T, U> {
     fn query_to_one(&self, query: &[T], index: usize) -> U {
         (self.metric)(query, &self.data[index])
     }
+
+    fn reorder(&mut self, indices: &Vec<usize>) {
```
nishaq503, line 135: Same type here.

src/core/dataset.rs
```diff
@@ -125,13 +131,49 @@ impl<T: Number, U: Number> Dataset<T, U> for VecVec<T, U> {
...
+                while source_index < i {
+                    source_index = indices[source_index];
+                }
+
+                // We swap to the correct index. Importantly, this index is always
+                // to the right of i, we do not modify any index to the left of i.
+                // Thus, because we followed the cycle to the correct index to swap,
+                // we know that the element at i, after this swap, is in the correct
+                // position.
+                self.data.swap(source_index, i);
```
nishaq503, line 166: Add a required method to the trait with the following signature:
```rust
fn swap(&mut self, i: usize, j: usize) -> Result<(), String>;
```

Then implement the method for this struct and then move the implementation of this method to be a default implementation is the trait itself.

nishaq503 (changes requested):
This is pretty good as a starting point. See the comments for the changes that I want. We can have a meeting to discuss these if you want.

src/core/dataset.rs
olwmc: For `VecVec`: What should the failure of a `swap` look like? The default behavior for `Vec::swap` when an index is out of bounds is to just panic. Should `VecVec::swap` do it's own bounds check `Err` accordingly or just let it panic?

nishaq503: It's fine to panic. You can change the signature to not return anything.

olwmc:
Requested changes have been made :)

src/core/dataset.rs
```diff
@@ -125,13 +132,55 @@ impl<T: Number, U: Number> Dataset<T, U> for VecVec<T, U> {
...
+
+                // We swap to the correct index. Importantly, this index is always
+                // to the right of i, we do not modify any index to the left of i.
+                // Thus, because we followed the cycle to the correct index to swap,
+                // we know that the element at i, after this swap, is in the correct
+                // position.
+                self.swap(source_index, i);
+            }
+        }
+    }
```
nishaq503, line 172: Move this implementation into the trait (at line 78), i.e. make it a default implementation.

src/core/dataset.rs
```diff
@@ -11,6 +11,7 @@ pub trait Dataset<T: Number, U: Number>: std::fmt::Debug + Send + Sync {
     fn indices(&self) -> Vec<usize>;
     fn one_to_one(&self, left: usize, right: usize) -> U;
     fn query_to_one(&self, query: &[T], index: usize) -> U;
+    fn swap(&mut self, i: usize, j: usize);
```
nishaq503, line 14: Add documentation to state the purpose of this method, and say that it is acceptable to panic if the indices are out of range.

nishaq503 (changes requested):
Really close now. Just a couple more changes are needed. See the new comments.

olwmc:
Changes made 😎

nishaq503:
Sweet! Append your info under `authors` in `Cargo.toml`, squash your commits down to one and let me know. Then I'll merge this.

olwmc:
Done! Thanks for teaching me a new git trick.

PR #99, opened 2023-04-03
Minor refactor for dataset

olwmc:
*Assigned to me by Najib*

# Overview
This PR contains a minor refactor, pulling the geometric median calculation logic out of `Cluster` and making it a method in the `Dataset` trait.

# Comments
Originally it was talked about moving the calculations for some of the other properties calculated in `Cluster::build` out into their own methods. I wasn't sure if this was still the case after reviewing the video from our original meeting. If this is the case, let me know I'll pull them out.

### DIFF
--- a/src/core/cluster.rs
+++ b/src/core/cluster.rs
@@ -166,17 +166,7 @@ impl<'a, T: Number, U: Number, D: Dataset<T, U>> Cluster<'a, T, U, D> {
             self.data.choose_unique(n, indices, self.seed)
         };
 
-        let arg_center = arg_samples[self
-            .data
-            .pairwise(&arg_samples)
-            .into_iter()
-            // TODO: Bench using .max instead of .sum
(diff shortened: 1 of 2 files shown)


nishaq503 (changes requested):
This looks good. Here are a couple of changes I want to see:

- Move the implementation of `median` into the trait itself. Then add to the documentation that the default implementation is not scalable (for now) to an arbitrarily large input.
- Add a panic condition to the docs for when `indices` is a zero-length slice.

Once these are done, the PR will be ready.

olwmc:
- [X] ~Move the implementation of median into the trait itself. Then add to the documentation that the default implementation is not scalable (for now) to an arbitrarily large input~
- [X] ~Add a panic condition to the docs for when `indices` is a zero-length slice.~

Finished. I wasn't sure what would be the best way to represent two notes stylistically, so I grouped them into a subheader (h2). This could be easily changed.

nishaq503 (approved):
This looks good. Squash the commits down to one and I'll merge it in.

olwmc:
This should be good 😎

PR #102, opened 2023-06-04
Levenshtein

morganprior:
Moved Levenshtein distance function over from distance crate. Altered function and  function signature to be consistent with existing edit distance function signatures.


ndaniels (approved):
Looks fine; tests still need improved.

nishaq503 (approved):
LGTM. We will update tests and benchmarks in a later PR

PR #104, opened 2023-06-04
Reordering logic

olwmc:


src/core/dataset.rs
```diff
@@ -151,24 +154,34 @@ pub trait Dataset<T: Number, U: Number>: std::fmt::Debug + Send + Sync {
             .unwrap()
             .0]
     }
+
+    // TODO: Clean up the names on these
+    fn set_reordered_indices(&mut self, indices: &[usize]);
+    // This method will fail if an `reorder` has not been called.
```
ndaniels, line 160: Anything that can fail should have a Result<T, E> (e.g. Result<(), E>) type

src/core/dataset.rs
```diff
@@ -151,24 +154,34 @@ pub trait Dataset<T: Number, U: Number>: std::fmt::Debug + Send + Sync {
...
+    fn get_reordered_index(&self, i: usize) -> usize;
```
ndaniels, line 161: This type should be Result<usize, E> for some error type E (which could be String, or Box<Error>)

ndaniels (commented):
Also, the return type for reorder() should be a Result type as well. Remember you can always use () ('unit', like void) to indicate that the value arm of the Result is never actually a value. This just forces the caller to verify that the call was successful, avoiding Hoare's billion dollar (now, likely trillion-dollar) mistake.

PR #108, opened 2023-06-10
Needleman wunsch

morganprior:
Implemented Needleman-Wunsch for CLAM.  @ndaniels @nishaq503 @EmilyL1019 

Some comments: 
- Alignments are currently being returned as a tuple of Vecs of Edits (Deletion(index) or Insertion(index, character) or Substitution(index, character)).  The first element in the tuple represents the edits needed to turn the first sequence in to the second one and the latter represents the edits needed to go in the other direction. This can easily be reverted to have the function return two aligned sequences (as opposed to the set of edits) if desired. 
-  I have a function which gets edit distance from the distance table, but this is not necessary; under the scoring scheme (match: 0, gap: 1, mismatch: 1), the edit distance is the same as levenshtein, so you might as well just use levenshtein if all you care about is the distance. 
- For the traceback, we have two versions: recursive and iterative. There are benchmarks for both in benches/needleman-wunsch.rs. Iterative seems significantly faster right now. 
- I have benchmarks for both kinds of traceback function as well as two for the overall process: one which varies dimensionality and one which varies size of the alphabet. 
- I've only written tests for a few hand-calculated examples, which pass. We still need more robust, less arbitrary testing. 
- Right now, the function only returns one alignment (i.e., ignores ties). I am working on an alternative version (in Sequence Analysis repo) which won't ignore ties, but I figured this version was good enough for our current purposes

### DIFF
--- a/benches/needleman-wunsch.rs
+++ b/benches/needleman-wunsch.rs
@@ -1,0 +1,104 @@
+use criterion::*;
+
+use clam::distances::alignment_helpers;
+use clam::distances::strings;
+use clam::utils::helpers;
+
+fn needleman_wunsch_dim(c: &mut Criterion) {
+    let mut group = c.benchmark_group("needleman-wunsch-dim");
+    group.significance_level(0.025).sample_size(10);
+
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
+    group.plot_config(plot_config);
+
+    group.sampling_mode(SamplingMode::Flat);
+
+    for n in [10, 25, 50, 100, 250, 500, 1000] {
+        let data = helpers::gen_data_u8(2, n, 65, 68, 42);
+        let x = &data[0];
+        let y = &data[1];
+        let id = BenchmarkId::new("nw", n);
+        group.bench_with_input(id, &n, |b, _| {
+            b.iter_with_large_drop(|| strings::needleman_wunsch_with_edits::<u8, u32>(x, y));
+        });
+    }
+
+    group.finish();
+}
+
+fn needleman_wunsch_alphabet_size(c: &mut Criterion) {
+    let mut group = c.benchmark_group("needleman-wunsch-alphabet-size");
+    group.significance_level(0.025).sample_size(10);
+
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
+    group.plot_config(plot_config);
+
+    group.sampling_mode(SamplingMode::Flat);
+
+    for alphabet_size in [4, 8, 16, 32] {
+        let data = helpers::gen_data_u8(2, 50, 65, 65 + alphabet_size, 42);
+        let x = &data[0];
+        let y = &data[1];
+        let id = BenchmarkId::new("nw", alphabet_size);
+        group.bench_with_input(id, &alphabet_size, |b, _| {
+            b.iter_with_large_drop(|| strings::needleman_wunsch_with_edits::<u8, u32>(x, y));
+        });
+    }
+
+    group.finish();
+}
+
+fn traceback_iterative_dim(c: &mut Criterion) {
+    let mut group = c.benchmark_group("traceback-iterative-dim");
+    group.significance_level(0.025).sample_size(10);
+
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
+    group.plot_config(plot_config);
+
+    group.sampling_mode(SamplingMode::Flat);
+
+    for n in [10, 25, 50, 100, 250, 500, 1000] {
+        let data = helpers::gen_data_u8(2, n, 65, 68, 42);
+        let x = &data[0];
+        let y = &data[1];
+        let table: Vec<Vec<(usize, alignment_helpers::Direction)>> = alignment_helpers::compute_nw_table(x, y);
+        let id = BenchmarkId::new("nw-traceback-iterative", n);
+        group.bench_with_input(id, &n, |b, _| {
+            b.iter_with_large_drop(|| alignment_helpers::traceback_iterative(&table, (x, y)));
+        });
+    }
+
+    group.finish();
+}
+
+fn traceback_recursive_dim(c: &mut Criterion) {
+    let mut group = c.benchmark_group("traceback-recursive-dim");
+    group.significance_level(0.025).sample_size(10);
+
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
+    group.plot_config(plot_config);
+
+    group.sampling_mode(SamplingMode::Flat);
+
+    for n in [10, 25, 50, 100, 250, 500, 1000] {
+        let data = helpers::gen_data_u8(2, n, 65, 68, 42);
+        let x = &data[0];
+        let y = &data[1];
+        let table: Vec<Vec<(usize, alignment_helpers::Direction)>> = alignment_helpers::compute_nw_table(x, y);
+        let id = BenchmarkId::new("nw-traceback-recursive", n);
+        group.bench_with_input(id, &n, |b, _| {
+            b.iter_with_large_drop(|| alignment_helpers::traceback_recursive(&table, (x, y)));
+        });
+    }
+
+    group.finish();
+}
+
+criterion_group!(
+    benches,
+    needleman_wunsch_dim,
+    needleman_wunsch_alphabet_size,
+    traceback_recursive_dim,
+    traceback_iterative_dim
+);
+criterion_main!(benches);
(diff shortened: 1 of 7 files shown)


src/distances/alignment_helpers.rs
```diff
@@ -0,0 +1,559 @@
...
+                column_index -= 1;
+            }
+            Direction::Left => {
+                aligned_x.insert(0, unaligned_x[column_index - 1]);
+                aligned_y.insert(0, T::from(b'-').unwrap());
+                column_index -= 1;
+            }
+            Direction::Up => {
+                aligned_x.insert(0, T::from(b'-').unwrap());
+                aligned_y.insert(0, unaligned_y[row_index - 1]);
```
nishaq503, line 108: Instead of `insert`ing at index `0`, push to the end of the vectors. Then, when returning from the function, reverse the vectors.

src/distances/alignment_helpers.rs
```diff
@@ -0,0 +1,559 @@
...
+fn _traceback_recursive<T: Number>(
+    table: &Vec<Vec<(usize, Direction)>>,
+    indices: (usize, usize),
+    unaligned_seqs: (&[T], &[T]),
+    aligned_seqs: (Vec<T>, Vec<T>),
+) -> (Vec<T>, Vec<T>) {
+    let (unaligned_x, unaligned_y) = unaligned_seqs;
+    let (mut aligned_x, mut aligned_y) = aligned_seqs;
+
+    let (mut row_index, mut column_index) = indices;
```
nishaq503, line 142: This function is likely slow because this is a clone happening because an immutable thing is now being made mutable. Instead, you should try initializing the vectors in the outer function, pass in mutable references to this function, and not have this function return anything.

src/distances/alignment_helpers.rs
```diff
@@ -0,0 +1,559 @@
...
+            column_index -= 1;
+        }
+        Direction::Left => {
+            aligned_x.insert(0, unaligned_x[column_index - 1]);
+            aligned_y.insert(0, T::from(b'-').unwrap());
+            column_index -= 1;
+        }
+        Direction::Up => {
+            aligned_x.insert(0, T::from(b'-').unwrap());
+            aligned_y.insert(0, unaligned_y[row_index - 1]);
```
nishaq503, line 160: Same as before about pushing to the end of the vectors and reversing at the end of the outer function

src/distances/alignment_helpers.rs
```diff
@@ -0,0 +1,559 @@
...
+        let y_u8 = "NAJIBEATSPEPPERS".as_bytes();
+        let table1: Vec<Vec<(usize, Direction)>> = compute_nw_table(x_u8, y_u8);
+        let nw_u8 = traceback_recursive(&table1, (x_u8, y_u8));
+        assert_eq!(
+            nw_u8,
+            (
+                [78, 65, 74, 73, 66, 45, 80, 69, 80, 80, 69, 82, 83, 69, 65, 84, 83].to_vec(),
+                [78, 65, 74, 73, 66, 69, 65, 84, 83, 80, 69, 80, 80, 69, 45, 82, 83].to_vec()
+            )
+        );
```
nishaq503, line 530: What do you see if you convert these types back to `str`? It's hard to intuitively recognize the answer when looking at a sequence of integers. Try doing the same for the rest of the tests.

src/distances/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub mod alignment_helpers;
```
nishaq503, line 1: Make this `pub(crate)` so that users of the library do not have access to these helpers.

src/distances/strings.rs
```diff
@@ -1,4 +1,5 @@
 use crate::core::number::Number;
+use crate::distances::alignment_helpers;
```
nishaq503, line 2: `use super::alignment_helpers;`

src/distances/strings.rs
```diff
@@ -43,9 +44,40 @@ pub fn levenshtein<T: Number, U: Number>(x: &[T], y: &[T]) -> U {
...
+    let table = alignment_helpers::compute_nw_table(x, y);
+    let (aligned_x, aligned_y) = alignment_helpers::traceback_recursive(&table, (x, y));
+
+    let edit_x_into_y = alignment_helpers::alignment_to_edits(&aligned_x, &aligned_y);
+    let edit_y_into_x = alignment_helpers::alignment_to_edits(&aligned_y, &aligned_x);
+
+    let edit_distance: usize = table[table.len() - 1][table[0].len() - 1].0;
+
+    ((edit_x_into_y, edit_y_into_x), U::from(edit_distance).unwrap())
+}
```
nishaq503, line 75: Add recursive and iterative variants of this, changing only line 67.

src/distances/strings.rs
```diff
@@ -43,9 +44,40 @@ pub fn levenshtein<T: Number, U: Number>(x: &[T], y: &[T]) -> U {
     }
 }
 
+// Wrapper function for calculating distance based on NW-aligned sequences (returns edit distance only)
+pub fn needleman_wunsch<T: Number, U: Number>(x: &[T], y: &[T]) -> U {
```
nishaq503, line 48: Add proper doc-strings for this function and add a link to an explanation of the distance function, e.g. an online tool or a Wikipedia page etc.

src/distances/strings.rs
```diff
@@ -43,9 +44,40 @@ pub fn levenshtein<T: Number, U: Number>(x: &[T], y: &[T]) -> U {
...
+}
+
+/*
+Function for calculating calculating distance based on NW-aligned sequences (returns edit
+distance AND alignment associated with shortest edit distance)
+
+For now, in cases where there exist ties for the shortest edit distance, we only return
+one alignment
+*/
+pub fn needleman_wunsch_with_edits<T: Number, U: Number>(
```
nishaq503, line 62: Add a proper doc-string here as well.

benches/needleman-wunsch.rs
```diff
@@ -0,0 +1,104 @@
...
+    let mut group = c.benchmark_group("needleman-wunsch-dim");
+    group.significance_level(0.025).sample_size(10);
+
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
+    group.plot_config(plot_config);
+
+    group.sampling_mode(SamplingMode::Flat);
+
+    for n in [10, 25, 50, 100, 250, 500, 1000] {
+        let data = helpers::gen_data_u8(2, n, 65, 68, 42);
```
nishaq503, line 17: Use a wider range of minimum and maximum values.

benches/needleman-wunsch.rs
```diff
@@ -0,0 +1,104 @@
...
+        let x = &data[0];
+        let y = &data[1];
+        let id = BenchmarkId::new("nw", n);
+        group.bench_with_input(id, &n, |b, _| {
+            b.iter_with_large_drop(|| strings::needleman_wunsch_with_edits::<u8, usize>(x, y));
```
nishaq503, line 22: Change the distance type to `u32`.

nishaq503 (changes requested):
This looks really good. There are a small number of changes that I want made, mostly in the `alignment_helpers`file.

src/distances/alignment_helpers.rs
```diff
@@ -0,0 +1,559 @@
...
+    let gap_penalty = 1;
+
+    // Initialize top row and left column distance values
+    for row in 0..(len_y + 1) {
+        table[row][0] = (gap_penalty * row, Direction::None);
+    }
+
+    for column in 0..(len_x + 1) {
+        table[0][column] = (gap_penalty * column, Direction::None);
+    }
```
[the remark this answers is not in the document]
morganprior: They should be. I originally thought you'd never hit those cells during traceback, but you can.

src/distances/alignment_helpers.rs
morganprior: Sounds good

src/distances/strings.rs
morganprior: What is the rationale behind this? I've seen things done in both ways throughout CLAM and never understood why.

src/distances/strings.rs
morganprior: Okay. I assume you mean with some sort of input flag like "recursive = TRUE"

benches/needleman-wunsch.rs
morganprior: Ok-- this was meant to be representative of ACGT sequences which is why I only did a 4 char alphabet

src/distances/strings.rs
nishaq503: For now, I just want two separate functions. This will be for benchmarks when you make the `alignment_helpers` module `pub(crate)`

nishaq503 (approved):
LGTM

PR #118, opened 2023-07-06
Updates to distance functions

nishaq503:
This PR removes all distance functions (except Needleman Wunch) from this repo and, instead, relies on our `distances` crate for the distance functions.

It does the same for the `Number` trait.

It also removes the `synthetic_data` util as we start to rely on the `SyMaGen` crate for synthetic data generation.


ndaniels (approved):
This looks good.

PR #120, opened 2023-07-09
Fixed bug with cluster indices.

nishaq503:
Also: 

* Improved the signature of the Dataset trait
* Improved the signature of the Cluster struct


ndaniels (approved):
Approving mystery fix. :-/

PR #122, opened 2023-07-10
Feature: search enums

nishaq503:
This PR makes it so that we now use enums of search algorithms for RNN and KNN search. It also updates the relevant benchmarks and examples, along with adding some better documentation in the README.


ndaniels (approved):
“Second class citizen” is deserved.

PR #129, opened 2023-07-30
Updates/separate search modules

morganprior:
Gave rnn and knn separate folders in preparation for adding new knn variants.


abd-clam/src/cakes/knn/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! K-Nearest Neighbors (KNN) variants.
+
+pub mod _knn;
```
nishaq503, line 3: Make this a private module. Then do a public export the enum `KnnAlgorithm`. It should look like:

```rust
mod _knn;

pub use _knn::KnnAlgorithm;
```

abd-clam/src/cakes/mod.rs
```diff
@@ -4,8 +4,8 @@ pub mod codec;
 pub mod knn;
 pub mod rnn;
 
-pub use knn::KnnAlgorithm;
-pub use rnn::RnnAlgorithm;
+pub use knn::_knn::KnnAlgorithm;
+pub use rnn::_rnn::RnnAlgorithm;
```
nishaq503, line 8: The above change will allow these use statements to remain the same as before

abd-clam/src/cakes/rnn/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! Ranged-Nearest Neighbors (RNN) variants.
+
+pub mod _rnn;
```
nishaq503, line 3: Same change as with `_knn`

nishaq503 (changes requested):
I need a couple of changes for how imports are happening. Otherwise, this is good.

nishaq503 (approved):
LGTM

PR #131, opened 2023-08-01
Update/variant-thresholds-no-separate-centers

morganprior:
- added new variant of KnnAlgorithm (thresholds without separate centers -- the branch name is misleading but the PR name is correct) 
- added test for new variant 
- added benchmark for new variant


morganprior:
Note: The version on my other branch (pre clippy changes) is faster. Not sure if that's to be expected but just noting.

nishaq503 (approved):
This looks good. Just bump the release version and we can merge this.

PR #132, opened 2023-08-02
Limited Arrow IPC support

olwmc:
## Overview
The main focus of this PR is `BatchedArrowDataset` which implements the `Dataset` trait and can read directories with one or more Arrow IPC files. The last of which (when sorted lexicographically) may have fewer instances than the preceding shards in the event of an uneven split.

Note that the Arrow files currently supported is only a small subset. Most notably we only take statically sized types and the files themselves are limited to one chunk each.

Notably this PR contains code that will be removed in the future. `ConstructableNumber` is a construct I needed that will likely get merged into some part of the `Number` trait from `distances`.

This also incorporates some changes made by Najib. Namely, the removal of the `Copy` trait being a requirement for `T` in `Dataset` and the consequences related to this change.

As this is a ground-up reimplementation of a narrow subset of the Arrow format. *I expect* there will be some bugs, especially when it comes to larger datasets and uneven splits. It is highly recommended to assert important qualities of the dataset are true (i.e. cardinality, dimensionality, etc.)  when using `BatchedArrowDataset` to avoid invalid results. I have not been able to produce incorrect results but my data is not necessarily representative.


olwmc:
My most recent commit has my `Dataset` type changes. Note that this still does not compile due to the issues regarding `Number::from_ne_bytes`

olwmc:
My previous commit (please for the love of god ignore my commit message mess up) implements a basic interface to assure that `T` can be constructed from btyes. I then use this trait in the dataset implementation and thus remove all existing errors. None of these changes are final. I just want *"how I'm wrapping my head around this"* out of my brain.

PR #133, opened 2023-08-07
Update/sieve v2

morganprior:
Added knn SieveV2 to cakes module along with associated test and adjustment to benches.


nishaq503 (approved):
LGTM

PR #135, opened 2023-08-10
Optimized sieve V1.

nishaq503:
In this PR, I have optimized SieveV1, remove the sieve struct altogether and arriving at a relatively clean single function for the search algorithm.

I have also added a wrapper around a priority queue, called `Hits`, to keep track of the `k` best hits during search.

I have also used this new `Hits` struct in linear search, which is now ~4.5x faster than before.

I have attached the current benchmarks (from my mac). The y-axis is `knn-{metric}/{knn::Algorithm}/{k}` and the x-axis is the mean time (in `ms` on a log-scale) for `100` queries.
knn-bench

### DIFF
--- a/abd-clam/src/cakes/knn/sieve_v1.rs
+++ b/abd-clam/src/cakes/knn/sieve_v1.rs
@@ -133,13 +133,6 @@ impl<'a, T: Send + Sync + Copy, U: Number> Grain<'a, T, U> {
         (self.multiplicity <= k) || self.is_leaf
     }
 
-    /// A Grain is "inside" the threshold if the furthest, worst-case possible point is at most as far as
-    /// threshold distance from the query.
-    #[allow(dead_code)]
-    pub fn is_inside(&self, threshold: U) -> bool {
-        self.d <= threshold
-    }
-
     /// A Grain is "outside" the threshold if the closest, best-case possible point is further than
     /// the threshold distance to the query.
     pub fn is_outside(&self, threshold: U) -> bool {
--- a/abd-clam/src/cakes/knn/sieve_v1.rs
+++ b/abd-clam/src/cakes/knn/sieve_v1.rs
@@ -37,17 +37,13 @@ where
     loop {
         // The threshold is the minimum distance, so far, which guarantees that
         // the k nearest neighbors are within the threshold.
-        let i = Grain::partition(&mut grains, k, hits.len());
+        let i = Grain::partition(&mut grains, k);
         let threshold = grains[i].d;
 
         // Remove hits which are outside the threshold.
         hits.pop_until(threshold);
 
         let (insiders, non_insiders) = grains.split_at_mut(i);
-        if non_insiders.is_empty() {
-            grains_to_hits(data, query, &mut hits, insiders);
-            break;
-        }
 
         let (small_insiders, insiders) = insiders
             .iter_mut()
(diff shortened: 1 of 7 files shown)


abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -291,13 +197,13 @@ impl<'a, T: Send + Sync + Copy, U: Number> Grain<'a, T, U> {
...
-    pub fn _partition(grains: &mut [Self], l: usize, r: usize) -> usize {
-        let pivot = (l + r) / 2;
+    fn _partition_once(grains: &mut [Self], l: usize, r: usize) -> usize {
+        let pivot = l + (r - l) / 2;
         grains.swap(pivot, r);
 
         let (mut a, mut b) = (l, l);
         while b < r {
-            if grains[b].d <= grains[r].d {
+            if grains[b].d < grains[r].d {
```
morganprior, line 206: I forgot about this until now but my original reason for having it this way is that it allows the index returned from partition_kth to eliminate the need for the is_outside() function in sieve V1. You just only take grains[0:i] and you have everything you need and don't need to check one by one if they're outside or not. But if you get rid of the = you get less swaps which is good but you forego the ability to eliminate a bunch of grains without iterating over all of them. Not sure if that's worth considering. Either way, if you're going to get rid of the = you should change the documentation on this function because it is no longer correct.

abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -291,13 +197,13 @@ impl<'a, T: Send + Sync + Copy, U: Number> Grain<'a, T, U> {
     /// Changes pivot point and swaps elements around so that all elements to left
     /// of pivot are less than or equal to pivot and all elements to right of pivot
     /// are greater than pivot.
-    pub fn _partition(grains: &mut [Self], l: usize, r: usize) -> usize {
-        let pivot = (l + r) / 2;
+    fn _partition_once(grains: &mut [Self], l: usize, r: usize) -> usize {
+        let pivot = l + (r - l) / 2;
```
morganprior, line 201: Are these two pivots not equivalent? 

l + (r - l)/2 = l + r/2 - l/2 = l/2 + r/2 = (l+r)/2

abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -29,257 +27,165 @@ where
...
-                .number
-                > threshold
-        {
-            self.hits
-                .pop_max()
-                .unwrap_or_else(|| unreachable!("`hits` is non-empty"));
+        // If there are no outsiders then the search is complete.
+        if outsiders.is_empty() {
+            grains_to_hits(data, query, &mut hits, non_outsiders);
+            break;
```
morganprior, line 51: I trust you but I don't yet understand this so leaving a comment here for me to think this through. How is this not activating on the first iteration? The root cluster won't be an outsider so you'll have no outsiders so need to think through why this isn't just hitting on the root cluster

abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -29,257 +27,154 @@ where
...
 /// A Grain is a structure which stores a cluster, a distance, and a multiplicity.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 struct Grain<'a, T: Send + Sync + Copy, U: Number> {
-    /// Just something we have to do.
-    t_: std::marker::PhantomData<T>,
     /// The cluster.
     c: &'a Cluster<T, U>,
     /// The distance of the cluster's center to the query.
     d: U,
```
morganprior, line 108: I think this documentation is incorrect because even though creating a new grain requires passing in the distance from the cluster's center to the query, that's not what d actually represents since it gets immediately set to d + radius and represents d + radius in the algorithm thereafter.

abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -29,257 +27,154 @@ where
...
+            .map(|&mut g| g)
+            .partition::<Vec<_>, _>(|g| g.is_small(k));
+        grains_to_hits(data, query, &mut hits, &small_insiders);
+
+        let (leaves, non_leaves) = non_insiders
+            .iter_mut()
+            .filter(|g| !g.is_outside(threshold))
+            .map(|&mut g| g)
+            .partition::<Vec<_>, _>(|g| g.is_leaf);
+
```
morganprior, line 63: See what happens if you use is_small() rather than is_leaf() here. Likely makes negligible difference but now that we are rewarded for having more points in hits, it may offer improvements

abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -29,257 +27,154 @@ where
...
-        let (mut insiders, straddlers): (Vec<_>, Vec<_>) = self
-            .grains
-            .drain(..)
-            .filter(|g| !Grain::is_outside(g, threshold))
-            .partition(|g| Grain::is_inside(g, threshold));
-
-        // Distinguish between those insiders we won't partition further and those we will
-        // Add instances from insiders we won't further partition to hits
-        let (small_insiders, big_insiders): (Vec<_>, Vec<_>) = insiders
+        grains = non_leaves
```
morganprior, line 71: In the same vein as my previous comment, what about just adding all the leaves to hits? I.e., no hit makes it to the next iteration. Very non-rigorously, if you're a cluster and you've survived long enough for us to partition you down to a leaf, you must not suck too much lest you'd have been eliminated by now.

abd-clam/src/cakes/knn/sieve_v1.rs
```diff
@@ -29,257 +27,154 @@ where
...
+        let i = Grain::partition(&mut grains, k, hits.len());
+        let threshold = grains[i].d;
+
+        // Remove hits which are outside the threshold.
+        hits.pop_until(threshold);
+
+        let (insiders, non_insiders) = grains.split_at_mut(i);
+        if non_insiders.is_empty() {
+            grains_to_hits(data, query, &mut hits, insiders);
+            break;
```
morganprior, line 49: move right after line 40 and just check if i is last index in grains

PR #136, opened 2023-08-17
Feats/improved sieve

morganprior:
* Added two new k-NN sieve variants wherein, rather than maintain a separate priority queue for hits, hits are treated as grains. 
* One variant treats centers of grains separately; the other does not.
* Changed partition method to make more informed choices of pivot based on mean cluster cardinality. (This made major speed improvements but we may be able to do even better. Pivots are very concentrated around 0 for early iterations and k for the last few iterations, and we don't really capitalize on the "near k for the last few iterations" part yet. All of my attempts to do so made it slower.) 
* Fixed bugs and edge-case handling in partition methods

Note: I did not change the partition method for SieveV1 and SieveV2 to implement the trick above, though it would likely improve them as well. Because these new versions were faster than Sieve V1 and Sieve V2 regardless, that didn't seem worth doing right now. 

Another note: I promise, this will be my last PR for a while while I go in my cave and write the paper!!


nishaq503 (approved):
LGTM!

PR #141, opened 2023-09-20
Feat/serde for cluster

olwmc:
Implements basic json serialization for `Cluster`. 

This interface is heavily subject to change in response to the needs imposed by deserializing an entire `Tree` show themselves. 

The struct `SerializedCluster` is used as an intermediate representation to avoid writing a json deserializer directly for `Cluster`. This is primarily due to the fact that we *do not* want to store the actual children of a cluster, just the cluster itself. I also make use of another intermediate structure `SerializedChildInfo` which contains information about the children (if applicable, other wise it's serialized as `None`) like `arg_l/r`, `polar_distance`, and the names of the children.

The second struct is returned by `SerializedCluster::into_partial_cluster` (the naming here is so, so bad) if and only if the original `Cluster` had children. This allows us to not have to store child info directly but also always recover A) a `Cluster` from a `SerializedCluster` and B) any child info if applicable. 

### TODO
- [x] Inverse `Cluster::name` to reconstitute history after deserialization
- [x] A few more basic tests

### DIFF
--- a/abd-clam/src/core/cluster/mod.rs
+++ b/abd-clam/src/core/cluster/mod.rs
@@ -10,6 +10,7 @@ mod _cluster;
 mod criteria;
 mod tree;
 
-pub use _cluster::Cluster;
+#[allow(clippy::module_name_repetitions)]
+pub use _cluster::{Cluster, SerializedChildInfo, SerializedCluster};
 pub use criteria::{PartitionCriteria, PartitionCriterion};
 pub use tree::Tree;
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.21.6
+current_version = 0.21.7
 commit = False
 tag = False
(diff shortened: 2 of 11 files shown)


olwmc:
This should be ready for review. I think the majority of the functionality we need to serialize a `Tree` in terms of what we need from `Cluster` is here. Naming, function signatures, etc. I'm not so sure about.

nishaq503 (approved):
LGTM

olwmc:
@nishaq503 Changes back up. [Actions passed](https://github.com/olwmc/clam/actions/runs/6292229713/job/17081425742). Take another pass over to make sure I didn't miss anything but I think we're good

abd-clam/src/core/cluster/mod.rs
```diff
@@ -10,6 +10,7 @@ mod _cluster;
 mod criteria;
 mod tree;
 
-pub use _cluster::Cluster;
+#[allow(clippy::module_name_repetitions)]
+pub use _cluster::{Cluster, SerializedChildInfo, SerializedCluster};
```
nishaq503, line 14: We will likely end up making `SerializedChildInfo` private within the `_cluster` module.

nishaq503 (approved):

PR #146, opened 2023-10-02
Feat/dataset save load

olwmc:
Implements `save` and `load` functionality for numeric `vec2d`. Serializes into a simple binary format.

This code is pretty grungy but it's mostly low level file io and byte parsing.

The file format is the obvious one you'd think of:
- `VEC2D` raw text header (Basic read protection. Basically no cost)
- Null-terminated dataset name string
- Cardinality as a u64
- Dimensionality as a u64
- Null-terminated `type_name` string as defined in the `Number` crate
- Row major data. Byte width of the type is NOT encoded. Type width (and thus parsing) is dictated by the constructed type of `V` (see second impl block for numeric-type restricted `vec2d)` May want to include a warning in the constructor. We don't have any kind of underlying type id system for numeric types, so there's not really a clean way of encoding the individual dimension type. This is analogous to the compromise made for the arrow dataset code. 
- A single boolean byte (1 or 0) corresponding to `self.reordering.is_some()`
- If the previous byte is 1, then what follows is a sequence of u64's corresponding to the reordering map

This PR adds `temp_dir` as a dev dependency.

### DIFF
--- a/abd-clam/src/core/dataset/vec2d.rs
+++ b/abd-clam/src/core/dataset/vec2d.rs
@@ -1,5 +1,11 @@
 //! A dataset of a Vec of instances.
 
+use std::{
+    fs::File,
+    io::{Read, Write},
+    path::Path,
+};
+
 use distances::Number;
 use rand::prelude::*;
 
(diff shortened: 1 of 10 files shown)


abd-clam/src/core/dataset/vec2d.rs
```diff
@@ -50,6 +56,163 @@ impl<T: Send + Sync + Copy, U: Number> VecDataset<T, U> {
...
+        let cardinality = self.data.len();
+        let cardinality_bytes = cardinality.to_le_bytes();
+        handle.write_all(&cardinality_bytes).map_err(|err| format!("{err}"))?;
+
+        // Write dimensionality
+        let dimensionality = self.data[0].len();
+        let dimensionality_bytes = dimensionality.to_le_bytes();
+        handle
+            .write_all(&dimensionality_bytes)
+            .map_err(|err| format!("{err}"))?;
```
nishaq503, line 86: Include the `type_name` after this. I just added the method to the `Number` trait. You'll have to update the dependency on `distances` to `1.4.1`.

abd-clam/src/core/dataset/vec2d.rs
```diff
@@ -50,6 +56,163 @@ impl<T: Send + Sync + Copy, U: Number> VecDataset<T, U> {
...
+        handle
+            .read_exact(&mut dimensionality_buf)
+            .map_err(|err| format!("{err}"))?;
+
+        // Convert them to usize
+        let cardinality = usize::from_le_bytes(cardinality_buf);
+        let dimensionality = usize::from_le_bytes(dimensionality_buf);
+
+        // Get each data point
+        let type_size = std::mem::size_of::<V>();
```
nishaq503, line 165: Change this in accordance with the `type_name` that you will have written. Ideally, you would check to see that the name matches and then use the `num_bytes` method on `Number`.

abd-clam/src/core/dataset/vec2d.rs
```diff
@@ -206,4 +368,64 @@ mod tests {
...
+
+    #[test]
+    fn test_save_load_deterministic() {
+        let data = vec![vec![1, 2, 3, 4, 5], vec![6, 7, 8, 9, 10]];
+        let data: Vec<&[u32]> = data.iter().map(Vec::as_slice).collect();
+
+        let mut dataset = VecDataset::new("test".to_string(), data, euclidean_sq::<u32, u32>, false);
+        let indices = dataset.indices().iter().map(|x| *x).rev().collect::<Vec<usize>>();
+        dataset.set_reordered_indices(&indices);
+        dataset.save(&std::env::temp_dir().join("datset.save")).unwrap();
```
nishaq503, line 380: Add and use the `tempdir` [crate](https://docs.rs/tempdir/latest/tempdir/) as a dev dependency.

nishaq503 (changes requested):
This looks very good. There are two changes that I'd like to see:

- Update the `distances` dependency to version `1.4.1` and then use the `type_name` method as part of the header.
- Use the `tempdir` [crate](https://docs.rs/tempdir/latest/tempdir/) as a dev-dependency for the tests.

olwmc:
Changes made. Random test changed a little bit to speed it up a little bit.

nishaq503 (dismissed):
LGTM

PR #147, opened 2023-10-06
Fixing workflows

nishaq503:
This PR:

- Updates github actions to run on pull-requests from forks.
- fixes clippy lints from Rust `v1.73` for implementations of `PartialOrd` and `Ord` traits
- for Rust tests, only runs tests in abd-clam because the other crates have their own workflows.
- for Python tests, skips the classifier tests from py-clam because that implementation was really hacky.


morganprior (approved):
LGTM!

PR #150, opened 2023-10-14
Removed automl submodule. Moved py-clam into submodule

nishaq503:


ndaniels:
I don't want to be a pain here, but I'd like a more detailed description of what's up with automl removal. Is this just reorganization?

nishaq503:
> I don't want to be a pain here, but I'd like a more detailed description of what's up with automl removal. Is this just reorganization?

I removed `automl` because I don't control the release of the package. Also, I'd rater move us to `smartcore`.

ndaniels (approved):
Ok, looks good.

PR #151, opened 2023-10-21
Feat: Save/load for tree

olwmc:
## TODO
- [x] Recover depth info
- [x] Overall cleanup
- [x] Docs for all new funcs
- [x] Comments
- [x] Basic randomized testing
- [x] General improvements; Removal of constant values where applicable

## Notes
- Might be able to cut significant time by keeping track of allocated clusters. Currently we just traverse which amounts to an expensive noop
- We should have a slightly more robust way of comparing equality between two cluster hierarchies 
- Do we ever plan to expand to parametrized cluster partitioning? I.e. trinary, quaternary, etc.
- **BUG**: `tree.depth` is always zero because we never actually recalculate it after partitioning, lol. Once that gets fixed we need to update the `assert_trees_equal` function.

### DIFF
--- a/abd-clam/src/core/cluster/_cluster.rs
+++ b/abd-clam/src/core/cluster/_cluster.rs
@@ -884,24 +884,24 @@ mod tests {
         let mut c1 = Cluster::new_root(&data, Some(42));
         c1.history = vec![true, true, false, false, true];
 
-        let (s1, s1_children) = SerializedCluster::from_cluster(&c1);
-        let s1_string = serde_json::to_string(&s1).unwrap();
+        let (original, original_children) = SerializedCluster::from_cluster(&c1);
+        let original_string = serde_json::to_string(&original).unwrap();
 
-        let s2: SerializedCluster = serde_json::from_str(&s1_string).unwrap();
-        assert_eq!(s1.name, s2.name);
-        assert_eq!(s1.seed, s2.seed);
-        assert_eq!(s1.offset, s2.offset);
-        assert_eq!(s1.cardinality, s2.cardinality);
-        assert_eq!(s1.arg_center, s2.arg_center);
-        assert_eq!(s1.arg_radial, s2.arg_radial);
-        assert_eq!(s1.radius_bytes, s2.radius_bytes);
-        assert_eq!(s1.lfd, s2.lfd);
-        assert_eq!(s1.ratios, s2.ratios);
+        let deserialized: SerializedCluster = serde_json::from_str(&original_string).unwrap();
+        assert_eq!(original.name, deserialized.name);
+        assert_eq!(original.seed, deserialized.seed);
+        assert_eq!(original.offset, deserialized.offset);
+        assert_eq!(original.cardinality, deserialized.cardinality);
+        assert_eq!(original.arg_center, deserialized.arg_center);
+        assert_eq!(original.arg_radial, deserialized.arg_radial);
+        assert_eq!(original.radius_bytes, deserialized.radius_bytes);
+        assert_eq!(original.lfd, deserialized.lfd);
+        assert_eq!(original.ratios, deserialized.ratios);
 
-        let c2 = s2.into_partial_cluster();
+        let c2 = deserialized.into_partial_cluster();
 
         assert_eq!(c1, c2);
-        assert!(s1_children.is_none());
+        assert!(original_children.is_none());
     }
 
     #[test]
--- a/abd-clam/src/core/cluster/_cluster.rs
+++ b/abd-clam/src/core/cluster/_cluster.rs
@@ -656,17 +656,11 @@ pub struct SerializedCluster {
     pub lfd: f64,
     /// The `Cluster`'s ratios
     pub ratios: Option<Ratios>,
-    // /// Serialized information about the cluster's immediate children, if applicable
-    // pub children: Option<SerializedChildren>,
 }
 
 /// Serialized information about a given `Cluster`'s children
 #[derive(Serialize, Deserialize)]
 pub struct SerializedChildren {
-    /// The encoded history of the left child
-    pub left_name: String,
-    /// The encoded history of the right child
-    pub right_name: String,
     /// The left pole of the `Cluster`
     pub arg_l: usize,
     /// The right pole of the `Cluster`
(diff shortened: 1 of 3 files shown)


abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+    /// # Errors
+    /// Errors out on any directory or file creation issues
+    #[allow(clippy::missing_panics_doc)]
+    pub fn save(&self, path: &Path) -> Result<(), String> {
+        // Create our directory if needed
+        let dirbuilder = DirBuilder::new();
+
+        if !path.exists() {
+            dirbuilder.create(path).map_err(|e| e.to_string())?;
+        }
```
nishaq503, line 105: Assume that the `user_given_path` already exists. Raise an error otherwise.

abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+    ///      dataset/    <-- The serialized dataset.
+    ///      leaves.json <-- A json file of leaf names.
+    /// ````
+    ///
+    /// # Errors
+    /// Errors out on any directory or file creation issues
+    #[allow(clippy::missing_panics_doc)]
+    pub fn save(&self, path: &Path) -> Result<(), String> {
+        // Create our directory if needed
+        let dirbuilder = DirBuilder::new();
```
nishaq503, line 101: What is the difference between `DirBuilder::create` and `File::create`?

abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+            if cur.is_leaf() {
+                leaves.push(filename.clone());
+            } else {
+                // Unwrapping is justified here because we validated that the cluster
+                // is a leaf before reaching this code
+                #[allow(clippy::unwrap_used)]
+                let [l, r] = cur.children().unwrap();
+                stack.push(l);
+                stack.push(r);
+            }
```
nishaq503, line 137: Use a `match` statement on `cur.children()` instead of this. That will remove the use of `unwrap` altogether.

abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+    }
+}
+
+/// Serializes a serializeable (`impl Serialize`) object to a given path
+///
+/// # Errors
+/// This function will error out on serialization or file i/o errors.
+fn serialize_to_file<S: Serialize>(path: PathBuf, object: &S) -> Result<(), String> {
+    let mut file = File::create(path).map_err(|e| e.to_string())?;
+    let info_string = serde_json::to_string(&object).map_err(|e| e.to_string())?;
```
nishaq503, line 255: Add a TODO here to encode the `object` as bytes instead of `json`. We'll add the byte-encodings later.

abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+///
+/// # Errors
+/// This function will error out on any deserialization or file i/o errors.
+fn recover_serialized_cluster<U: Number>(path: PathBuf) -> Result<Cluster<U>, String> {
+    let mut buffer = String::new();
+    let mut cluster_handle = File::open(path).map_err(|e| e.to_string())?;
+
+    cluster_handle.read_to_string(&mut buffer).map_err(|e| e.to_string())?;
+
+    let cluster: SerializedCluster = serde_json::from_str(&buffer).map_err(|e| e.to_string())?;
```
nishaq503, line 269: Add a complementary TODO here to decode from bytes.

abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+/// This function will error out on any deserialization or file i/o errors.
+fn recover_serialized_childinfo(path: PathBuf) -> Result<SerializedChildren, String> {
+    let mut buffer = String::new();
+    let mut childinfo_handle = File::open(path).map_err(|e| e.to_string())?;
+
+    childinfo_handle
+        .read_to_string(&mut buffer)
+        .map_err(|e| e.to_string())?;
+
+    let childinfo: SerializedChildren = serde_json::from_str(&buffer).map_err(|e| e.to_string())?;
```
nishaq503, line 285: Same complementary TODO here

abd-clam/src/core/cluster/tree.rs
```diff
@@ -72,4 +78,372 @@ impl<I: Instance, U: Number, D: Dataset<I, U>> Tree<I, U, D> {
...
+    /// reconstructed tree.
+    ///
+    /// # Errors
+    /// This function will return an error if there's any file creation, file reading, or json deserialization issues.
+    #[allow(clippy::missing_panics_doc)]
+    pub fn load(path: &Path, metric: fn(&I, &I) -> U, is_expensive: bool) -> Result<Self, String> {
+        // Alises to relevant directories
+        let cluster_path = path.join("clusters");
+        let childinfo_path = path.join("childinfo");
+        let dataset_dir = path.join("dataset").join("data");
```
nishaq503, line 171: Add a check for each of these to make sure the paths exist.

abd-clam/src/core/cluster/_cluster.rs
```diff
@@ -894,7 +884,7 @@ mod tests {
         let mut c1 = Cluster::new_root(&data, Some(42));
         c1.history = vec![true, true, false, false, true];
 
-        let s1 = SerializedCluster::from_cluster(&c1);
+        let (s1, s1_children) = SerializedCluster::from_cluster(&c1);
         let s1_string = serde_json::to_string(&s1).unwrap();
 
         let s2: SerializedCluster = serde_json::from_str(&s1_string).unwrap();
```
nishaq503, line 890: Rename `s1` and `s2` to indicate which came from the in-memory cluster and which was created from the json-string.

nishaq503 (changes requested):
I need a few changes here, most of which I have put in comments. There are also some implementation changes I can see that will make this faster. I'll experiment with some on my own and we can talk about them after.

abd-clam/src/core/cluster/tree.rs
olwmc: Couldn't find a reason to use `DirBuilder` directly, but it seems like `std::fs::create_dir` is the standard way to create directories.

olwmc:
Oop my bad. Last commit should read "Fixed `Tree` not updating its depth after partition". Forgot to escape the \`

nishaq503 (approved):
LGTM

PR #156, opened 2023-11-04
Moved tests outside of `src`

nishaq503:
This PR:

- Moves tests outside of individual files in `src` into their own standalone module.
- Makes the `Cluster` and some related structs public (in a limited fashion).
  - I always meant to do this and was waiting on something relatively stable. This looks like a good time. We may make more things public later.


abd-clam/README.md
```diff
@@ -12,44 +12,40 @@ CLAM is a library crate so you can add it to your crate using `cargo add abd_cla
 ### Cakes: Nearest Neighbor Search
 
 ```rust
-use symagen::random_data;
-
 use abd_clam::{knn, rnn, Cakes, PartitionCriteria, VecDataset};
 
-/// Euclidean distance function.
+/// The distance function with with do perform clustering and search.
```
ndaniels, line 17: not gonna reject a PR for a typo in a comment, but fix this.

ndaniels (approved):
Looks fine. Found a typo. Fix it later.

PR #161, opened 2023-11-09
Implemented Search trait

nishaq503:
This PR implements a `Search` trait and turns the `Cakes` struct into an enum with minimal changes in usage patterns.

This new trait will let us experiment with several strategies for making shards and benchmark those strategies using the common API provided by the Search trait.

The user-facing changes are minimal as can be seen by the lack of changes in `benches` and the simplifications in `tests`.

This PR also fixes recall calculations for the binaries in `cakes-results` and some tweaks to plot generation for the paper.

TODO:

- [x] Bump the version before merging.
- [x] Separate scaling-plots by `k`.

### DIFF
--- a/abd-clam/README.md
+++ b/abd-clam/README.md
@@ -68,7 +68,7 @@ let criteria = PartitionCriteria::default();
 
 // The Cakes struct provides the functionality described in the CHESS paper.
 // We use a single shard here because the demo data is small.
-let model = Cakes::new_single_shard(data, Some(seed), &criteria);
+let model = Cakes::new(data, Some(seed), &criteria);
 // This line performs a non-trivial amount of work. #understatement
 
 // At this point, the dataset has been reordered to improve search performance.
--- a/abd-clam/README.md
+++ b/abd-clam/README.md
@@ -12,7 +12,7 @@ CLAM is a library crate so you can add it to your crate using `cargo add abd_cla
 ### Cakes: Nearest Neighbor Search
 
 ```rust
-use abd_clam::{knn, rnn, Cakes, PartitionCriteria, VecDataset};
+use abd_clam::{knn, rnn, PartitionCriteria, Search, SingleShard, VecDataset};
 
 /// The distance function with with to perform clustering and search.
 ///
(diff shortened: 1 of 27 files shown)


abd-clam/README.md
```diff
@@ -67,8 +67,9 @@ let data = VecDataset::<Vec<f32>, f32>::new(
 let criteria = PartitionCriteria::default();
 
 // The Cakes struct provides the functionality described in the CHESS paper.
-// This line performs a non-trivial amount of work. #understatement
```
morganprior, line 70: I missed this documentation before lol

abd-clam/src/cakes/mod.rs
```diff
@@ -1,6 +1,340 @@
...
+    }
+
+    /// Creates a new CAKES instance with a randomly sharded dataset.
+    ///
+    /// # Arguments
+    ///
+    /// * `shards` - The shards of the dataset to search.
+    /// * `seed` - The seed to use for the random number generator.
+    /// * `criteria` - The criteria to use for partitioning the tree.
+    #[must_use]
```
morganprior, line 46: Why is this here?

morganprior (commented):
Looks good. I have some questions for you about the overall paradigm of implementing something like Search or Cluster as a trait, but I'll go over those in person.

ndaniels:
> Looks good. I have some questions for you about the overall paradigm of implementing something like Search or Cluster as a trait, but I'll go over those in person.

I have questions too. I worry that this could be an anti-pattern. Cakes into an enum seems fine, but to my mind the point of a trait -- which is like a typeclass in Haskell -- is to express a promise that certain behavior will be (sensibly) implemented, and often so that it can be used in lieu of the static type system for dynamic dispatch. Just like I tend to not like when C++ classes do all the work in their constructor (I would rather see a `do_work()` method) I want to be more clearly convinced that this is really making things more ergonomic or enabling some flexibility that we need. I want to look this over more, or chat with Najib, before approving.

abd-clam/src/cakes/mod.rs
nishaq503: Clippy

abd-clam/README.md
nishaq503: I only recently added it.

abd-clam/src/cakes/mod.rs
ndaniels: The idea is that a pure function (one with no side effects such as I/O) is pointless if its return value isn't used, so Clippy suggests you put in that attribute. That attribute causes the compiler to issue a warning if you ever call the function without using its return value. It does NOT mean that code must call that function.

ndaniels (approved):
Overall this looks good.

ndaniels (approved):

PR #163, opened 2023-11-11
Parallelized distance calculations for expensive metrics

nishaq503:
Also added a getter for shards in Cakes.

TODO:

- [x] Bump version before merge.


morganprior (approved):
Maybe rename `is_old` but other than that, LGTM!

PR #165, opened 2023-11-11
Added Save/Load methods to Cakes.

nishaq503:
See commit messages for summary of changes.

This update will allow for the super-expensive builds of trees to be saved and reused.

TODO:

- [x] Bump version

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.25.2
+current_version = 0.25.3
 commit = False
 tag = False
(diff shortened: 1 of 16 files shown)


ndaniels (approved):
This looks good. I want to add logging so that we have some way of communicating progress. This is not as easy as it was for compressive genomics, because there we could just report how many datapoints had been handled. I know several ways to do this, it should NOT hold up this PR, but I can work on it tomorrow.

morganprior (approved):
Looks good!

djperrone (approved):
This functionality should work well with my visualization. Did some preliminary testing and didn't run into any issues.

PR #167, opened 2023-11-14
Monorepo

thoward27:


ndaniels (commented):
Why is the rust toolchain being included in this? I would much rather use my rustup-maintained rust toolchain. This feels far too much like python .venv stuff. I'm OK with a monorepo, but *not* ok with requiring virtual environments. That's fine to work around python ecosystem failings, but shouldn't be necessary with cargo.

thoward27:
> Why is the rust toolchain being included in this? I would much rather use my rustup-maintained rust toolchain. This feels far too much like python .venv stuff. I'm OK with a monorepo, but _not_ ok with requiring virtual environments. That's fine to work around python ecosystem failings, but shouldn't be necessary with cargo.

Hey @ndaniels, are you talking about integrating Rust through `hermit`? If so, I'd be interested in discussing it with you. I believe `hermit` could bring significant value to the team. I'd liken it more to `pyenv` than `virtualenv`, although that comparison isn't perfect. Similar to `pyenv`, `hermit` lets you work with multiple versions of a given toolchain. However, unlike `pyenv`, there's no global configuration or binary sharing between repos and it lets you include a number of packages beyond just a language runtime. 

If we adopt hermit, the team would only need to have hermit installed to dive right in. Plus, it ensures everyone is consistently on the correct version of the toolchain. 

Beyond that, anyone can opt-out of using hermit if they want (though as far as I can see there is no benefit to not using it).

What are your thoughts on this? Would you be interested in hopping on a short call to play around with `hermit` together and discussing it as a part of our toolchain?

ndaniels:
Yeah, I generally don’t like adding layers of complexity. I have generally avoided pyenv and my view is any rust developer should have a rust environment that is consistent across projects. So I would rather not bundle rust with hermit. Perhaps hermit could be used for other things, especially any Python stuff since Python lacks a single consistent tool chain. All my software development experience has led me to be leery of adding complexity - even when it seems to simplify things into a turnkey approach.I’d be happy to do a call, but not until after Thanksgiving — too swamped right now. On Nov 17, 2023, at 4:39 PM, Tom Howard ***@***.***> wrote:﻿

Why is the rust toolchain being included in this? I would much rather use my rustup-maintained rust toolchain. This feels far too much like python .venv stuff. I'm OK with a monorepo, but not ok with requiring virtual environments. That's fine to work around python ecosystem failings, but shouldn't be necessary with cargo.

Hey @ndaniels, are you talking about integrating Rust through hermit? If so, I'd be interested in discussing it with you. I believe hermit could bring significant value to the team. I'd liken it more to pyenv than virtualenv, although that comparison isn't perfect. Similar to pyenv, hermit lets you work with multiple versions of a given toolchain. However, unlike pyenv, there's no global configuration or binary sharing between repos and it lets you include a number of packages beyond just a language runtime.
If we adopt hermit, the team would only need to have hermit installed to dive right in. Plus, it ensures everyone is consistently on the correct version of the toolchain.
Beyond that, anyone can opt-out of using hermit if they want (though as far as I can see there is no benefit to not using it).
What are your thoughts on this? Would you be interested in hopping on a short call to play around with hermit together and discussing it as a part of our toolchain?

PR #170, opened 2023-11-17
Added logging and improved performance of some distance calculations.

nishaq503:
The PR:

- Adds logging to the cluster building methods. Logs are optional and only show up if configured in a binary.
- Improved performance of the `pairwise` distances function in `Dataset` to do half the work.


ndaniels (approved):
Looks good.

PR #171, opened 2023-11-17
Added support for metadata in the `VecDataset` struct.

nishaq503:
Also:

- Added example for getting metadata of search results in the README.
- Implemented the `Instance` trait for all `Number` types for use as metadata. For example, one can use:
  - `bool`, `u16` or any integer type for labels.
  - `f32` and `f64` as scores in regression.
  - `String` for headers of sequences.
  - Any arbitrary type that implements `Instance` can be used as well.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.25.5
+current_version = 0.26.0
 commit = False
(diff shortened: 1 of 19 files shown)


ndaniels (approved):
Looks good. I still really dislike the generic term "Instance" because it's too much a term of art in OOP already.

PR #186, opened 2024-01-21
Fix: Cluster partition for non-metrics no longer has infinite recursion

nishaq503:
I found a rare logical error that only shows up when the distance function is not a metric, specifically when it does not have an identity among points, i.e. when `dist(x, x) > 0`. In such cases, when a cluster was partitioned, sometimes all points would be assigned to a single child instead of being divided among the two children.

This PR fixes the issue by checking if all points have been assigned to the same child and short-circuiting `partition`.

This PR also includes some minor fixes to functions in `utils` and makes `utils` a public module for use in companion crates.

TODO:

- [x] Merge [this PR](https://github.com/URI-ABD/clam/pull/187) first.


.github/workflows/tests.yml
nishaq503, line 1: This file changed is handled in the PR linked in the description of this PR

ndaniels (approved):
Looks good. Holding off on merging.

morganprior (approved):
LGTM!

PR #193, opened 2024-01-25
feat: Scoring functions for graph

andrewmlefebvre:


crates/abd-clam/src/core/graph/_graph.rs
```diff
@@ -198,7 +198,7 @@ pub struct Graph<'a, U: Number> {
 }
 
 impl<'a, U: Number> Graph<'a, U> {
-    /// Creates a new `Graph` from the provided `tree`.
+    // Creates a new `Graph` from the provided `tree`.
```
nishaq503, line 201: Bring this doc-string back.

crates/abd-clam/src/core/graph/_graph.rs
```diff
@@ -795,8 +795,8 @@ mod tests {
             let selected_clusters = select_clusters(
                 raw_tree.root(),
                 &pretrained_models::get_meta_ml_scorers().first().unwrap().1,
-                i,
-            );
+            )
+            .unwrap();
```
nishaq503, line 799: Don't use `unwrap` here. Use the `?` operator

crates/abd-clam/src/core/graph/criteria.rs
```diff
@@ -3,25 +3,103 @@
...
-    for c in root.subtree() {
-        if c.depth() == depth || (c.depth() < depth && c.is_leaf()) {
-            selected_clusters.insert(c);
-        }
+    scoring_function: &MetaMLScorer,
+) -> Result<ClusterSet<'a, U>, String> {
+    let mut cluster_set: HashSet<&'a Cluster<U>> = HashSet::new();
+    let mut clusters = score_clusters(root, scoring_function)?;
+
+    clusters.retain(|item| item.cluster.depth() > 3);
```
nishaq503, line 91: This will remove any leaves at a lower depth. You need to keep such leaves around. Also, the minimum depth should be an input parameter.

nishaq503 (changes requested):
This is pretty good,; only minor changes are needed:

- The `select_clsuters` function should have a `min_depth` input parameter instead of defaulting to `4`. I defaulted to `4` for the CHAODA paper abut that was hand-wavey.
- the `criteria.rs` file has a few functions/structs that only ever used within that file. Such things should not be made public to other modules.

crates/abd-clam/src/core/graph/_graph.rs
andrewmlefebvre: Since this is a test function that does not return an error you can not use the ? to propagate the error

crates/abd-clam/src/core/graph/criteria.rs
andrewmlefebvre: Fixed

crates/abd-clam/src/core/graph/_graph.rs
andrewmlefebvre: Fixed

djperrone:
lgtm

nishaq503 (approved):

PR #197, opened 2024-02-03
added tiebreaker to cluster comparison. Changed partition back to ret…

djperrone:
fix: added tiebreaker to cluster comparison. Changed partition back to 
retain after benchmarking


nishaq503 (approved):
LGTM!

PR #198, opened 2024-02-06
Added random vector generation within n-dim sphere

morganprior:
Implemented function to generate a random point (in Cartesian coordinates) from a uniform distribution over an n-dimensional ball of a given radius. The math used is derived from [this wikipedia article](https://en.wikipedia.org/wiki/N-sphere#Spherical_coordinates). The algorithm here is linear-time in the dimensionality of the point being generated.

Some comments: 
- We've had errors with functions in Symagen before, so I think we should think about a better way to test this stuff going forward (I know this is difficult because of the randomness) 
-  Right now this is implemented for f64s; is that okay or should it be more generic?


nishaq503 (approved):
LGTM!

morganprior (commented):
LGTM!

PR #200, opened 2024-02-13
Updating existing chaoda code to be clippy compliant

andrewmlefebvre:


andrewmlefebvre:
Although this code now passes linter tests, I would assume there might still be comments on stylistic choices on how that was done.  Feel free to leave comments on if there are cleaner ways to simplify some of these changes.

crates/abd-clam/src/chaoda/graph_scorers.rs
```diff
@@ -4,6 +4,7 @@ use std::hash::Hash;
 use crate::core::cluster::Cluster;
 use crate::core::graph::Graph;
 use distances::Number;
+use statistical::{mean, population_standard_deviation};
```
nishaq503, line 7: There is a `utils` module in `abd-clam` crate and it provides functions to calculate mean, variance and std-dev. Use those instead of the `statistical` crate. Once done, move `statistical` back to a `dev-dependency` in `Cargo.toml`.

crates/abd-clam/Cargo.toml
```diff
@@ -56,6 +56,8 @@ libm = "0.2.7"
 # memmap2 = "0.8.0"
 smartcore = "0.2.1"
 
+statistical = "1.0.0"
+
```
nishaq503, line 60: See my comment in `graph_scorers.rs` about this.

nishaq503 (changes requested):
I have added a comment about the `statistical` crate. Other than that, I need to see some tests added for the restored functionality.

andrewmlefebvre:
Not sure what happened with the last commit.  Will raise a new PR soon to cleanup changed files.  Any notable changes needed from actual changed files before doing so?

PR #202, opened 2024-02-16
Added python wrapper for distances crate

nishaq503:
Along with docs, tests, benchmarks and a README with plots.

TODO:

- [x] Update Earthly setup to run python tests in CI


ndaniels (approved):
This looks good.

.pre-commit-config.yaml
```diff
@@ -30,18 +30,29 @@ repos:
         exclude: '.bumpversion.cfg'
       - id: check-merge-conflict
 
-  - repo: https://github.com/psf/black
```
thoward27, line 33: Removing black without adding another formatter is concerning

Earthfile
```diff
@@ -1,7 +1,30 @@
-VERSION 0.7
-FROM rust:latest
+VERSION 0.8
+FROM ubuntu:latest
```
thoward27, line 2: I'd love to think through keeping this base image as `rust` and using `python` for some of the python targets. Not that that would be the best way, I just feel like it has some advantages and would be simpler.

Earthly allows each target to inherit from whatever base image it needs, and it can pull files from any other image to do its work.

Earthfile
```diff
@@ -15,17 +38,27 @@ RUN cargo install \
...
+RUN python3.9 -m venv .venv
+RUN . .venv/bin/activate && pip install --upgrade pip
+
 # This target prepares the recipe.json file for the build stage.
 chef-prepare:
-    COPY --dir crates Cargo.toml .
+    COPY --dir crates .
+    COPY --dir python .
+    COPY Cargo.toml .
     RUN cargo chef prepare
```
thoward27, line 52: Does chef prepare actually use any of the python code to do its work?

ruff.toml
```diff
@@ -0,0 +1,78 @@
+select = [
+  "F",   # pyflakes
```
thoward27, line 2: May want to `select = [ "ALL" ]` then ignore the ones you don't like.

thoward27 (commented):
Overall this looks mergable, but I think it would be good to discuss the changes to the Earthfile and to CI together. IMO inheriting from multiple targets feels like it would be cleaner than producing an ubuntu image capable of doing everything

Earthfile
nishaq503: That would be cool. We still need some unified tests for the whole workspace and compiling the python wrappers will require a python interpreter to be present

Earthfile
nishaq503: I don't know. However, not copying the python folder means that the rust workspace cannot be compiled because of missing members

ruff.toml
nishaq503: Done

PR #205, opened 2024-02-19
Feat/triangle search

djperrone:


crates/abd-clam/src/core/graph/_graph.rs
```diff
@@ -757,11 +758,112 @@ impl<'a, U: Number> Graph<'a, U> {
...
+
+    /// Finds triangles in the graph.
+    ///
+    /// A triangle consists of three clusters where each cluster is connected to the other two.
+    ///
+    /// # Returns
+    ///
+    /// Returns a vector of tuples, where each tuple represents a triangle found in the graph.
+    /// Each tuple contains references to three clusters forming the triangle.
+    pub fn find_triangle_clusters(&self) -> Vec<(&'a Cluster<U>, &'a Cluster<U>, &'a Cluster<U>)> {
```
nishaq503, line 770: Change the return type to `Vec<[&'a Cluster; 3]>`

crates/abd-clam/src/core/graph/_graph.rs
```diff
@@ -757,11 +758,112 @@ impl<'a, U: Number> Graph<'a, U> {
...
+    /// Finds triangles in the graph and returns them as tuples of edges.
+    ///
+    /// A triangle consists of three clusters where each cluster is connected to the other two.
+    /// This function finds triangles in the graph and returns them as tuples of edges.
+    ///
+    /// # Returns
+    ///
+    /// Returns a vector of tuples, where each tuple represents a triangle found in the graph.
+    /// Each tuple contains references to three edges forming the triangle.
+    pub fn find_triangle_clusters_to_edges(&'a self) -> Vec<(&'a Edge<U>, &'a Edge<U>, &'a Edge<U>)> {
```
nishaq503, line 802: This function feel unnecessary

crates/abd-clam/src/core/graph/_graph.rs
```diff
@@ -928,4 +1032,161 @@ mod tests {
 
         true
     }
+
+    #[test]
+    fn test_triangles() {
```
nishaq503, line 1037: Move these tests to a new file in the `tests` directory.

crates/abd-clam/src/core/graph/_graph.rs
```diff
@@ -7,6 +7,7 @@ use distances::Number;
 // use crate::core::{cluster::Cluster, dataset::VecDataset};
 use crate::core::graph::criteria::{detect_edges, select_clusters};
 
+use crate::VecDataset;
```
nishaq503, line 10: Why is this import here? I would prefer all methods to be generic over all datasets

nishaq503 (changes requested):
I have added some specific comments among the files. Beyond those, I have some overarching comments:

- For the `find_triangle_clusters` function, I believe this approach will bias the quality metrics to preserving local structure. We should balance this out by choosing some number of random combinations of clusters with which to make triangles. Using clusters that do not share edges should allow us to measure how well we preserve global structure.
- Given the above point, we should not need a function to find the edges that make a triangle. However, even if we do not *need* such a function, if it would improve runtime performance of calculating the quality metrics, we should keep the functions. This requires adding to the `benches` directory

PR #211, opened 2024-04-25
Dice Distance

andrewmlefebvre:


crates/distances/src/sets/mod.rs
```diff
@@ -46,3 +46,40 @@ pub fn jaccard<T: Int, U: Float>(x: &[T], y: &[T]) -> U {
...
+///
+/// let distance: f32 = dice(&x, &y);
+///
+/// assert!((distance - 0.5).abs() < f32::EPSILON);
+/// ```
+pub fn dice<T: Int, U: Float>(x: &[T], y: &[T]) -> U {
+    if x.is_empty() || y.is_empty() {
+        return U::one();
+    }
+    let intersection_size = U::from(x.iter().filter(|v| y.contains(v)).count());
```
nishaq503, line 77: This operation is quadratic-time in the lengths of `x` and `y`. You should first convert `x` and `y` into `HashSet`s and then use the relevant methods to determine the size of the intersection. That will make this operation linear-time instead

nishaq503 (changes requested):

nishaq503 (approved):
LGTM

PR #213, opened 2024-06-25
Kulsinski distance, set tests added

brianmaterne:
I coded in Kulsinksi distance in the same file as Jaccard distance using the same format, relatively self explanatory.
I also added a test_sets.rs file to the tests folder, which contains tests for the Jaccard and Kulsinski distances and can easily be added to. I use random exhaustive testing, creating two 1000 length sets with each number randomly swapped to be on or off. I then check if each function gives the intended result finding distance between the two.
I then have some checks of edge cases, with one or both sets being empty or both sets being the same.
Both of these testing methods are easily addable to once a new set distance function is added.
I also added and checked off Kulsinski distance in the ReadMe.

I also added Dice distance into the ReadMe, and added Dice distance into my tests when moving my code from distances to clam. Even though there were already Dice tests, I added Dice into my own as well in order to be encompassing.

### DIFF
--- a/crates/distances/src/sets/mod.rs
+++ b/crates/distances/src/sets/mod.rs
@@ -91,8 +91,14 @@ pub fn dice<T: Int, U: Float>(x: &[T], y: &[T]) -> U {
 /// Kulsinski distance.
 ///
 /// Similar to the Jaccard distance, the Kulsinski distance is a measure of the dissimilarity
-/// between two sets. It is defined as the cardinality of the intersection divided by the
-/// sum of the number of not equal dimensions and the total number of dimensions.
+/// between two sets. It is defined as the sum of the number of not equal dimensions and the
+/// total number of dimensions minus the number of elements in the intersection, all divided by
+/// the sum of the number of not equal dimensions and the total number of dimensions.
+///
+/// # Links
+///
+/// <https://scikit-learn.org/stable/modules/generated/sklearn.metrics.DistanceMetric.html>
+/// <https://docs.scipy.org/doc/scipy-1.7.1/reference/reference/generated/scipy.spatial.distance.kulsinski.html>
 ///
 /// # Arguments
 ///
--- a/crates/distances/src/sets/mod.rs
+++ b/crates/distances/src/sets/mod.rs
@@ -87,3 +87,46 @@ pub fn dice<T: Int, U: Float>(x: &[T], y: &[T]) -> U {
 
     U::one() - (U::from(2) * intersection_size / size)
 }
+
+/// Kulsinski distance.
+///
+/// Similar to the Jaccard distance, the Kulsinski distance is a measure of the dissimilarity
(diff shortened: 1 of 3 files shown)


crates/distances/src/sets/mod.rs
```diff
@@ -87,3 +87,46 @@ pub fn dice<T: Int, U: Float>(x: &[T], y: &[T]) -> U {
 
     U::one() - (U::from(2) * intersection_size / size)
 }
+
+/// Kulsinski distance.
+///
+/// Similar to the Jaccard distance, the Kulsinski distance is a measure of the dissimilarity
+/// between two sets. It is defined as the cardinality of the intersection divided by the
+/// sum of the number of not equal dimensions and the total number of dimensions.
```
morganprior, line 95: Please add a link to your source for the definition of Kulsinski distance between merging. There are some different definitions out there. For example, [this definition from Scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.kulczynski1.html#scipy.spatial.distance.kulczynski1) is different from what you have. [This one](https://docs.scipy.org/doc/scipy-1.7.1/reference/reference/generated/scipy.spatial.distance.kulsinski.html) is an equivalent definition to what you have, but it is in terms of Boolean vectors, and it takes a bit of algebra to verify they are equivalent.

crates/distances/tests/test_sets.rs
```diff
@@ -27,3 +29,92 @@ fn test_dice() {
...
+
+/// Random exhaustive testing of set distances, manually creating union and intersection values
+#[test]
+fn sets_test() {
+    for _ in 0..10000 {
+        let x: Vec<u16> = gen_set();
+        let y: Vec<u16> = gen_set();
+        let mut union: f32 = 0.0;
+        let mut intersection: f32 = 0.0;
+        let mut size: f32 = 0.0;
```
morganprior, line 52: Not a big deal, but I think it's usually better to give these variables types that make sense based on what they store, and then cast to a different type when the need arises. For example, `union`, `intersection`, and `size` will always be non-negative integers less than 2000, so a usize type would suffice. 

The usize type will cause you problems when you start doing division, but you can get around that by doing something like f32::from(intersection) at that point, rather than preemptively making those variables f32s.

morganprior (changes requested):
Looks good, but you should add a link to a source before we merge.

nishaq503 (changes requested):
~After you've made the changes that @morganprior requested, you should bump the version of the distances crate. Use a "minor" bump. The command should be something line `bump2version minor`. See near the end of `CONTRIBUTING.md` for some more info and reach out if you have questions.~

brianmaterne:
I'll add a link to the source, where in the code should I put it? Just in the comment description of the function? I can also implement the other changes, those sound good.

nishaq503 (approved):
LGTM!

morganprior (approved):

PR #214, opened 2024-06-25
Feat/compression

morganprior:
Revised string distance functions, compression, compressed r-NN search, and _linear_ compressed k-NN search. All compressed search algorithms pass tests.


nishaq503 (changes requested):
This will need lots more work in the future, as we have already discussed. For now, we need some version bumps before merging. Specifically, "minor" bumps for `abd-clam`, `distances` and `SyMaGen`.

morganprior:
MERGE!!!!!

nishaq503 (approved):
LGTM!

PR #215, opened 2024-07-01
fix(release): reorder crates

thoward27:
I think the failure to publish on master comes down to not having run `distances` publish step first. This PR re-orders the CI matrix such that `distances` is published before we attempt to publish other packages.


nishaq503 (approved):
LGTM!

PR #216, opened 2024-07-05
Cleanup of the mono-repo

nishaq503:
This PR:

- standardizes the metadata files (LICENSE, .gitignore, etc.) in each crate,
- removes leftover files from when we switched to the mono-repo structure,
- properly uses workspace dependencies instead of managing dependency versions in each member crate
- removes the `pyo3` dependency from `crates/symagen` in preparation for `pypi/symagen`
- removes `crates/py-clam` in preparation for `pypi/abd-clam`


morganprior (approved):
LGTM!

PR #222, opened 2024-10-11
Rewrite for PanCAKES

nishaq503:
It's pretty much a complete rewrite.

CHAODA training is still broken and I will revisit that later.


ndaniels (approved):
LGTM!

ok more seriously this will take some time to fully digest but I don’t want to hold things up. Chaoda training should be fairly high priority.

Some regression tests for chaoda mml training might be a nice idea.

PR #226, opened 2024-12-04
Added MSA and separated Metric from Dataset

nishaq503:
This PR also contains the current code for reproducing the CAKES benchmarks for our submission to SIMODS after the first round of reviews.


ndaniels (approved):
This all looks good. I see no blocking issues; obviously there is a lot of room to still improve documentation and avoid magic numbers in code.

PR #230, opened 2025-05-08
Implemented Hausdorff distance function

KaiMaffucci:
Implemented the Hausdorff meta-distance function, generic over sets of points. I also wrote tests for identity, symmetry, and triangle inequality. 

IMPORTANT NOTE: I programmed Hausdorff to accept two sets of *points*, since this is how Hausdorff works as a meta-distance function. This means I couldn't use the gen_set() function to generate test sets, and instead relied on test cases with sets that I had developed and tested myself. If there is something you want me to do to make up for the less randomness, please let me know.

### DIFF
--- a/crates/distances/src/sets/mod.rs
+++ b/crates/distances/src/sets/mod.rs
@@ -153,11 +153,10 @@ where
     C: Fn(U, U) -> bool,          // function to compare two distances
     F: Fn(&[T], &[T]) -> U,       // function to calculate distance between two points
 {
-
     // note: using x and y is kind of not a good idea for variable names: I'll replace them if I think of something better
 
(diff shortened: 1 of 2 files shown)


ndaniels (changes requested):
The implementation looks fine.  However, I would like at least some more comprehensive testing, particularly something more generative. Using the rand crate, you should be able to generate vectors of random values of the appropriate length, and you could also generate longer instances. I'm not sure why you had trouble with the generating test data; we can discuss on slack.

KaiMaffucci:
I added random sets for the test case, but let me know if you want them to be more specific.

nishaq503 (approved):
LGTM!

PR #236, opened 2025-09-11
chore(shell): implement search function

thoward27:


nishaq503 (changes requested):
Looks great over all. I want the following changes:

- Have the user provide an output directory and save search results as json/yaml files in that directory. I'm thinking for each query, for each alg, a list of the indices of and distances to the neighbors.
- Add an example to the README

nishaq503 (approved):
LGTM!

PR #237, opened 2025-09-11
Implemented Pearson distance function

camilleuri:
Pretty much done as far as I can tell, though I'd appreciate feedback on whether or not I should add more tests.


nishaq503 (changes requested):
This looks great! The only change I want to see is for you to move the tests to their own function. The `lp_f32` test function is meant for the minkowski family of norms only.

camilleuri:
Thanks for the feedback Najib, I've moved the tests over.

nishaq503 (approved):
LGTM!

PR #244, opened 2026-03-27
cluster locations, profiling, musals, shell, and dependencies

nishaq503:
This PR:

- de-copules `clam` from `distances` and `symagen` so that we can move those crates to their own repositories in a subsequent PR. This will make it much simpler for reviewers to access and use the `clam` repo.
- moves `shell` from being a separate crate to instead being an optional feature, `clam-shell` in the `abd-clam` crate.
- adds a `NamedAlgorithm` trait to ease conversions to/from string representation of all algorithms so that they can be used in the CLI and, eventually, over a web interface.
- adds an `AlignedSequence` type to efficiently store gaps to allow scaling MSAs to 10M+ sequences
- changes clusters from being stored in a `HashMap` to be in a `Vec` indexed by their centers. This removes the hashing overhead of cluster look-up in search algorithms. This, by itself, improves search throughput by 2-20% depending on the algorithm, dataset, and metric.
- further optimizes search algorithms through targeted profiling.
- formalizes some approximate variants of knn-search algorithms. They all rely on a single `tol` (tolerance) hyper-parameter to control the throughput-vs-recall tradeoff.
- adds a `tqdm`-based progress bar to the terminal when building a tree.
- significantly improves documentation for `Tree`,  `Cluster`, `PartitionStrategy`, etc.

Some things that will come in subsequent PRs include:

- removal of `distances` and `symagen` from this repo after they get moved into their own repositories.
- in `clam-shell`:
  - bringing back the `pancakes` sub-command. Pancakes is tricky because I want to avoid exposing certain types and methods to the user so that I can remove the ability to break compression invariants, i.e. how items are compressed/decompressed in terms of each other. Once done, this will let us remove several error conditions and `Result` types from that module.
  - finishing a subcommand under `cakes` to allow exploring the throughput-vs-recall frontier for the approximate search algorithms.
  - adding a subcommand under `cakes` to augment datasets to benchmark algorithms with higher cardinality multipliers.
  - bringing back the `generate-date` subcommand, though this is a pretty low priority.


ndaniels (approved):
After lengthy discussion in person, I'm fine with this.

